PR #32, opened 2013-09-08
Changed managers patching 

Stranger6667:
I propose a solution to the problem, which I mentioned here https://github.com/jakewins/django-money/pull/28.
It's based on class_prepared signal usage.

### DIFF
--- a/djmoney/tests/model_tests.py
+++ b/djmoney/tests/model_tests.py
@@ -7,7 +7,7 @@ Created on May 7, 2011
 from django.test import TestCase
 from moneyed import Money
 from testapp.models import (ModelWithVanillaMoneyField,
-    ModelRelatedToModelWithMoney, ModelWithChoicesMoneyField, BaseModel, InheritedModel)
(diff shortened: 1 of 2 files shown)


inureyes:
@benjaoming This code has a potential problem when model uses two or more managers. This code overwrites every manager to MoneyManager, which causes critical errors.

e.g. If model has CurrentSiteManager as a field, 
```
on_site = CurrentSiteManager()
```
``make migrations`` command generates the migration code to alter on_site manager to MoneyManager.

benjaoming:
@inureyes that sounds reasonable, not sure if it's supposed to be so. Current code `setattr(sender, name, money_manager(manager))` looks like it's supposed to alter a manager, not entirely overwrite it. Maybe the fix should be done in `money_manager(manager)` ? You're most welcome to do a PR, I think we have a reasonable test level to ensure that if tests are green, your code is good...

inureyes:
@benjaoming Ok. I'll make a quick fix for that.

inureyes:
https://github.com/django-money/django-money/pull/194

Stranger6667:
I think we should cover this migration issue in tests. I can add them later, thank you for the report :)

PR #87, opened 2014-06-24
Fixing issue with model instances being given "objects" attribute

### LINKED ISSUE #86: Instances of models with MoneyFields have "objects" attribute
On any model I give a MoneyField, instances have an `objects` attribute, referring to that model's manager (as if it were the class instead). Seems to be a result of calling `patch_managers()` in `models/fields.py`.

ghinch:
Adding the "objects" attribute to model instances. Instead, replacing
the concrete_managers attribute on the model meta. This fixes #86.


Stranger6667:
Unfortunately it works only for Django < 1.8

PR #96, opened 2014-11-03
Add support for django-denorm

ErwinJunge:
This adds support for django-denorm. 

The problem it solves is that during conversion to sql pre_save is called on all the fields of the model by django, in order of addition to the class. The way django-denorm works is by overriding the pre_save on the denormalized field and calling the function to compute the denormalized value to set that instead.

Since in the case of a MoneyField the CurrencyField is set by the MoneyField, this doesn't work since the CurrencyField's pre_save method is called before the denormalized MoneyField, leading to XYZ in the database.

This PR calls the pre_save method of the denormalized MoneyField if django-denorm is in use to work around this issue.

The problem with this fix is that the denormalization function is called twice because of this, leading to extra computation (potential performance issue). Ideas to improve are welcome.


benjaoming:
Sorry about not looking at this before.

I would suggest the following:

1. Create your own field type extending MoneyField
2. Add the pre_save there

Would that work? I guess its already what you've done?

PR #104, opened 2015-01-08
Fixes #103: Data truncated for column

### LINKED ISSUE #103: Data truncated for column
It appears that the `Decimal.get_db_prep_save` is not called within `MoneyField.get_db_prep_save` if the input is a Money instance.

This works fine in sqlite (which the tests are based upon) but in MySQL if you supply a value with more decimal places that what the field supports the following is raised;

`_mysql_exceptions.Warning: Data truncated for column 'name' at row 8`

(I'll be submitting a patch shortly)

alexhayes:


benjaoming:
Thanks for a well-implemented fix!

PR #108, opened 2015-02-11
Sum or subtract between different currencies

dnmellen:
I really missed the ability to perform operations between different currencies and I modified `MoneyPatched` on `fields.py` to convert the currency to the local one in sum and sub operations.

    In [5]: (MoneyPatched(1, 'EUR') + Money(1, 'USD') - Money(1, 'USD') + Money(1, 'GBP')).amount
    Out[5]: Decimal('2.35')

You need to have installed django-money-rates to perform the conversion between currencies.

Disclaimer: I made some tests but there are some more pip requirements (mock)


benjaoming:
This looks really cool! And useful! And with auto-conversion of currencies and a test case.

I'm digging this. I'll try to merge it with the current master.

PR #115, opened 2015-04-22
Add Python 3 trove classifier

dekkers:
This adds the Python3 trove classifier so that PyPI and tools like caniusepython3 can see that Python 3 is supported.


benjaoming:
Thanks for this!

PR #114, opened 2015-04-07
Support for Django 1.8

pedrorodriguesgomes:

Issue #101, opened 2015-01-01
Appends _currency to non-money ExpressionFields


AlexRiina:
It seems like `_expand_money_params` [processes all ExpressionNodes (e.g. Q and F) as if they are MoneyFields which leads to querying for `$field_currency`](https://github.com/jakewins/django-money/blob/master/djmoney/models/managers.py#L36). I had some trouble getting all of the tests to pass (and your travis-ci widget points to the wrong project), but the below should evidence the error.

```python
diff --git a/djmoney/tests/model_tests.py b/djmoney/tests/model_tests.py
index 974be95..d062853 100644
--- a/djmoney/tests/model_tests.py
+++ b/djmoney/tests/model_tests.py
@@ -10,7 +10,7 @@ from .testapp.models import (ModelWithVanillaMoneyField,
     ModelRelatedToModelWithMoney, ModelWithChoicesMoneyField, BaseModel, InheritedModel, InheritorModel,
     SimpleModel, NullMoneyFieldModel, ModelWithDefaultAsDecimal, ModelWithDefaultAsFloat, ModelWithDefaultAsInt,
     ModelWithDefaultAsString, ModelWithDefaultAsStringWithCurrency, ModelWithDefaultAsMoney, ModelWithTwoMoneyFields,
-    ProxyModel)
+    ProxyModel, ModelWithNonMoneyField)
 import moneyed
 
 
@@ -171,6 +171,15 @@ class RelatedModelsTestCase(TestCase):
         ModelRelatedToModelWithMoney.objects.get(moneyModel__money__lt=Money("1000.0", moneyed.ZWN))
 
 
+class NonMoneyTestCase(TestCase):
+
+    def testAllowExpressionNodesWithoutMoney(self):
+        """ allow querying on expression nodes that are not Money """
+        ModelWithNonMoneyField(money=Money(100.0), desc="hundred").save()
+        instance = ModelWithNonMoneyField.objects.filter(desc=F("desc")).get()
+        self.assertEqual(instance.desc, "hundred")
+
+
 class InheritedModelTestCase(TestCase):
     """Test inheritence from a concrete model"""
 
diff --git a/djmoney/tests/testapp/models.py b/djmoney/tests/testapp/models.py
index 8a91882..aeb64c6 100644
--- a/djmoney/tests/testapp/models.py
+++ b/djmoney/tests/testapp/models.py
@@ -51,6 +51,11 @@ class ModelWithChoicesMoneyField(models.Model):
     )
 
 
[...]

PR #116, opened 2015-05-06
Fix tox environment

edwinlunando:
add latest pymoneyed to tox environment. cc @benjaoming


benjaoming:
Oh cool, I'll be checking back in shortly to see how this fares.. could you also merge in #114 so we can see how it tests out with tox configured for 1.8 ?

edwinlunando:
guys, the test is not compatible with Django 1.8. Django has dropped django.test.simple.

benjaoming:
@edwinlunando have you merged in #114 ? I don't know how much it takes care of...

edwinlunando:
@benjaoming alright, the travis has run. there are some errors on the test. I will check it later.

edwinlunando:
@benjaoming the test for Django 1.8 will be different from Django 1.4 and 1.5. Do we want to drop the support or use python six? I recommend we only support the current supported Django version.

benjaoming:
@edwinlunando that's not a recommendation I back up... the only way to allow people to run flexible systems and have painless transitions between different django versions is to release django applications that are compatible with several versions of django.

edwinlunando:
allright, I will use Python six then.

benjaoming:
Awesome :)

edwinlunando:
@benjaoming while I'm working on the test, do you mind looking into this error. I think it is a bug https://travis-ci.org/django-money/django-money/jobs/61447801

benjaoming:
@edwinlunando no sorry, my involvement only goes as far as to review and merge stuff :) I maintain, I don't dev :)

briankung:
@edwinlunando have you seen https://github.com/django-money/django-money/issues/101? **"Appends _currency to non-money ExpressionFields"**

PR #123, opened 2015-06-06
Django 1.8 support with updated tests

willhcr:
Building on the earlier work by @pedrorodriguesgomes for Django 1.8 support I've updated the test suite code to run under Django 1.7 and 1.8. I also expanded the CI configuration to cover 1.7 and 1.8 as well as Python 3.4. Lastly, I made some minor updates to the README to mention the support for these.


runtests.py
```diff
@@ -4,9 +4,28 @@
 import sys
 from django.conf import settings
 
+try:
+    from django import setup
```
benjaoming: This looks a bit odd :)

I'd suggest

```python
# Detect if django.db.migrations is supported
from django import VERSION as DJANGO_VERSION
NATIVE_MIGRATIONS = (DJANGO_VERSION >= (1, 7))
```

runtests.py
```diff
@@ -40,13 +55,21 @@
...
 
-# If you use South for migrations, uncomment this to monkeypatch
-# syncdb to get migrations to run.
-from south.management.commands import patch_for_test_db_setup
-patch_for_test_db_setup()
+# Native migrations are present in Django 1.7+
+# This also requires initializing the app registry with django.setup()
+# If native migrations are not present, initialize South and configure it for running the test suite
+if NATIVE_MIGRATIONS:
+    setup()
```
benjaoming: `from django import setup`
(because I removed it above)

benjaoming:
Are we adding migrations or something? Because it looks like some migration issues are being addressed even though the app currrently doesn't have any :)

benjaoming:
Otherwise looks great, thanks for making all this work finally with tests and all!!

willhcr:
I've added in your recommended changes. Also, the changes to migrations were to avoid this fatal error when testing with Django 1.7+:

```
RuntimeError: South does not support Django 1.7 or higher. Please use native Django migrations.
```

benjaoming:
@willhcr I get it, there are tests that ensure that django-money is compliant with other application's migrations, therefore it's needed... sorry, that was me not getting the point :)

Issue #126, opened 2015-08-08
Broken readonly_fileds representation for some fields in Django==1.8


vmspike:
in `./djmoney/__init__.py:53`
AdminReadonlyField.contents = djmoney_contents

This line brokes readonly_fileds representation for some fields like CharField with 'choices' (raw value instead of choice description), DateTimeField (not localizable, no timezone) etc.
Cause of this is not up to date function `djmoney_contents()` which is use `smart_unicode()` instead of `display_for_field()`.
I've not found necessary functional for djmoney in this function, so I suppose to remove it, or update if it's required for compatibility with previous versions of Django.

benjaoming:
Not quite sure I fully understand your description. But please do post a PR for this, sounds problematic!

Stranger6667:
This code was removed recently, so issue should be fixed now.

PR #127, opened 2015-08-09
Broken readonly_fileds representation for some fields in Django==1.8

vmspike:
Duplicate of issue #126:

in `./djmoney/__init__.py:53`
AdminReadonlyField.contents = djmoney_contents

This line brokes readonly_fileds representation for some fields like CharField with 'choices' (raw value instead of choice description), DateTimeField (not localizable, no timezone) etc.
Cause of this is not up to date function `djmoney_contents()` which is use `smart_unicode()` instead of `display_for_field()`.
I've not found necessary functional for djmoney in this function, so I suppose to remove it, or update if it's required for compatibility with previous versions of Django.

This pull request completely clean `./djmoney/__init__.py`, if you still need content of this file you should update function `djmoney_contents()` inside according to `django.contrib.admin.utils.AdminReadonlyField.contents()`, because I'm pretty sure it is original file `djmoney_contents()` was written from.


benjaoming:
This introduces a regression in how money fields will be displayed I suppose. But if it breaks the admin as you say, it would also warrant the merge.

For the future, do this:

1. Write a test case that breaks and push it to a PR. Then everyone can see what breaks.
2. Create the fix in a second commit.

benjaoming:
Closing this but hoping and waiting for a solution without regressions.

PR #133, opened 2015-08-31
Better check on None value in decompress

tsouvarev:
In limist/py-moneyed@03768f6205e0829c98f0497938dca79ccd335743 was added `Money.__nonzero__`, which evaluates to `bool(self.amount)`.
So, if you have `default` argument for `MoneyField` with `amount` equal to 0, then `decompress` will return `[None, None]`, which is look like there is no `default` at all. 
If you change this `if` to check `None`-ness of value, then old behavior remains

### DIFF
--- a/djmoney/forms/widgets.py
+++ b/djmoney/forms/widgets.py
@@ -28,7 +28,7 @@ class MoneyWidget(MultiWidget):
         super(MoneyWidget, self).__init__(widgets, *args, **kwargs)
 
     def decompress(self, value):
-        if value:
+        if value is not None:
             return [value.amount, value.currency]
         return [None, None]
 


benjaoming:
Would this impose an update of the version dependency?

https://github.com/django-money/django-money/blob/master/setup.py#L32
https://github.com/django-money/django-money/blob/master/tox.ini#L24

benjaoming:
Btw thanks for the patch!

tsouvarev:
@benjaoming Well, I have no other problems with latest `py-moneyed`

benjaoming:
@tsouvarev what I meant was: You've added something that depends on a change in py-moneyed, does it have to be specified? Currently, we have > 0.4 (that's either 0.5 or 0.6).

So you're saying this works with 0.5 and 0.6 ?

tsouvarev:
@benjaoming My patch doesn't depend on `py-moneyed==0.6`, it's just too generic `if` become a bug after new version was released. You can safely apply patch and use any version of `py-moneyed`

benjaoming:
Okay I was just wondering because the commit you've referenced was only released in py-moneyed 0.6.

benjaoming:
Thanks for explaining!

tsouvarev:
Thanks! Will you update package on PyPI?

benjaoming:
I don't have the access, I've given @reinbach a mention in the latest commit where I bumped the version and tagged the release (0.7.2).

PR #134, opened 2015-09-19
Compare different currencies correctly -- fix #12

### LINKED ISSUE #12: "Cannot compare Money with different currencies" error because of lack of `__eq__` method
Django-money check that the currency are the same before most operations:

E.G, L91 classes.py:

        if self.currency == other.currency:
            return Money(
                amount=self.amount + other.amount,
                currency=self.currency)

        raise TypeError('Cannot add or subtract two Money ' +
                        'instances with different currencies.')

But because the Currency doesn't declare any `__eq__` method, the check is implemented by Python this way:

    self.currency == other.currency 

is turned into

    id(self.currency) == id(other.currency)

It works most of the time, as by importing a currency from the same module, you DO use the same object, and therefor it DOES have the same ID:

    from moneyed import USD

But It won't work if you are comparing instances that have been created in one interpreter, and compared in another, like in the case where you use queue managers like Celery.

There is an easy fix: just declare an `__eq__` method giving an universal way of comparing currencies:

    def __eq__(self, other):
        return self.code == other.code

While waiting for the patch, people that encounter this bug can monkey patch their own moneyed lib dynamically by adding this in THEIR app's `__init__.py`:

    from moneyed import Currency

    def __eq__(self, other):
        return self.code == other.code

    Currency.__eq__ = __eq__

benjaoming:
#12

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -103,6 +103,14 @@ class MoneyPatched(Money):
         return MoneyPatched._patch_to_current_class(
             super(MoneyPatched, self).__mul__(other))
 
+    def __eq__(self, other):
+        if hasattr(other, 'currency'):
+            if self.currency == other.currency:
+                return self.amount == other.amount
+            raise TypeError('Cannot add or subtract two Money ' +
+                            'instances with different currencies.')
+        return False
+
     def __truediv__(self, other):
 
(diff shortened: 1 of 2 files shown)


benjaoming:
Merging this, since tests are passing -- and there are quite a lot of tests about equality of Money types.

PR #136, opened 2015-09-20
Fix #97 - use choices setting for currency choices

### LINKED ISSUE #97: MoneyField currency_choices confusion
Current MoneyField `__init__` accepts both `choices` and `currency_choices` keyword args.

`choices`, however is completely ignored (not withstanding a line that assigns to it - the value is never used).

On the other hand `currency_choices` is used (it is passed to the `ChoiceField`, but if you use it, you get a warning saying that it is deprecated!

It's not obvious to me what the intended behaviour should be. Either it should issue the warning for `choices` not `currency_choices`, or it should actually use `choices`.

benjaoming:
#97


djmoney/forms/fields.py
```diff
@@ -17,17 +18,34 @@ class MoneyField(MultiValueField):
...
             warn('currency_choices will be deprecated in favor of choices', PendingDeprecationWarning)
             choices = currency_choices
-        decimal_field = DecimalField(max_value, min_value, max_digits, decimal_places, *args, **kwargs)
-        choice_field = ChoiceField(choices=currency_choices)
-        self.widget = currency_widget if currency_widget else MoneyWidget(amount_widget=decimal_field.widget,
-                                                                          currency_widget=choice_field.widget)
-        fields = (decimal_field, choice_field)
+
+        amount_field = DecimalField(max_value, min_value, max_digits, decimal_places, *args, **kwargs)
+        currency_field = ChoiceField(choices=choices)
```
benjaoming: This is where the fix happened, before it was still doing `choices=currency_choices`

benjaoming:
@spookylukey finally :)

PR #140, opened 2015-10-15
Fix field has changed

### LINKED ISSUE #112: _has_changed state incorrect
If you're just changing currency and leave the amount part intact, ``_has_changed`` returns False, which is not what one would expect.

plumdog:
Fixes #112 (I believe)

Changing the currency but leaving the amount the same was not considered a change, meaning that it wasn't being saved - I've added a test for this.

Had to do some tweaks, as the existing solution had been implemented so that inline forms in formsets (eg in admin) worked correctly when empty - there was an existing test for this.

Edit: pushed just failing test, waiting for travis to start for each environment. Will push fix when(?!) failing test fails on travis.

### DIFF
--- a/djmoney/forms/fields.py
+++ b/djmoney/forms/fields.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 from warnings import warn
 
 from django.core import validators
+from django.core.exceptions import ValidationError
 from django.forms import MultiValueField, DecimalField, ChoiceField
 from moneyed.classes import Money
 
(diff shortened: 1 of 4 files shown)


benjaoming:
Ah yes, Travis will only test the latest commit when pushing several at the same time. Awaiting the cure!

plumdog:
@benjaoming Cure applied.

benjaoming:
Thanks for adding tests, and thanks for a long-awaited fix for this issue!

PR #141, opened 2015-10-15
Fix tests for python2.6 and run on travis

plumdog:
`tox` was failing even once I had installed `python2.6`. So now - I think - tox should either pass or give you a `InterpreterNotFound`.

I also added two more environments for travis. Based on:
> Django 1.6 will be the final release series to support Python 2.6; beginning with Django 1.7, the minimum supported Python version will be 2.7.

I've only added these tests for Django 1.5 and 1.6.

(from https://docs.djangoproject.com/en/1.8/releases/1.6/#python-compatibility)

Also, there are no tests for Django 1.4 run on travis. (Though perhaps we don't want too many tests run on travis, in which case maybe my change here adding more tests is bad?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ install: pip install tox
 script: tox -e $ENV
 matrix:
   include:
+    - python: "2.6"
+      env: ENV=py26-django15-pm05
+    - python: "2.6"
+      env: ENV=py26-django16-pm05
     - python: "2.7"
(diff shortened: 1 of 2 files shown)


benjaoming:
That's really nice! I don't think django 1.4 is very necessary to support anymore. If people run super old versions of django, they might as well run super-old versions of django-money, too :)

benjaoming:
Thanks! :+1:

PR #143, opened 2015-10-22
fix loaddata: handle ignorenonexistent option

jack-cvr:


benjaoming:
Thanks for this!

Could you have a look at the failing django 1.5 and 1.6 tests..?

benjaoming:
Tests fixed, merging!

PR #147, opened 2015-11-01
Sort CURRENCY_CHOICES by name then by code #146

### LINKED ISSUE #146: Zambian Kwacha causes migrations.AlterField every makemigrations
The Zambian Kwacha causes migrations.AlterField every makemigrations because the order of choices does not remain on subsequent runs

This seems to have been caused by https://github.com/limist/py-moneyed/commit/56e79ef557432bdc86363e7fc6beb6598a47cd62

django-money (0.7.3)
py-moneyed (0.6.0)

synotna:
As CURRENCY_CHOICES may have multiple currency codes with the same name, we must sort by name then code otherwise the order is not reliable which then causes problems with django migrations

Fixes #146

### DIFF
--- a/djmoney/settings.py
+++ b/djmoney/settings.py
@@ -19,4 +19,4 @@ else:
     CURRENCY_CHOICES = [(c.code, c.name) for i, c in CURRENCIES.items() if
                         c.code != DEFAULT_CURRENCY_CODE]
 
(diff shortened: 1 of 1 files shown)


benjaoming:
That makes sense and passes tests!

PR #157, opened 2016-01-06
Fix for get_or_create / create manager methods not respecting currenc…

toudi:
…y code.

For instance, when you would do:

`mymodel, created = Model.objects.get_or_create(money_currency='EUR')`

and your model would have default_currency set to XYZ then the mymodel
instance would have the currency set to XYZ, despite the fact that you
have set it to EUR. This has to do with the fact that the currency field
has a lower creation counter than the basic money field itself.

This patch solves this problem.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -1,4 +1,5 @@
 from __future__ import division
+from django import VERSION
 from django.db import models
 from django.conf import settings
 try:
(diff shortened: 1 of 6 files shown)


benjaoming:
@reinbach after merging this PR, django-money should have django 1.8 and django 1.9 support, would you be kind to bump the version and do another release? :)

reinbach:
@benjaoming bumped version and released! :)

benjaoming:
Hurray!!!

Ship often

PR #162, opened 2016-01-06
fixes #161 by adding correct paths for py.test discovery

### LINKED ISSUE #161: Wrong pattern in pytests.ini ?
Report from @toudi :

> well, my point is that if you look into the djmoney/tests directory, there are several files with tests. However, they are skipped during the tests execution because in the pytest.ini there curently is only one entry:
> 
> python_files=mone*.py
> 
> which means that only one file fits this regexp: money_patched.py
> 
> That's why I asked whether this is a desired entry, or we should add another entries so that the other files get picked up, or we could also rename all the tests/*.py files to match money prefix ;)

Needs investigation!

benjaoming:
#161

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,5 @@
 [pytest]
-python_files=mone*.py
+python_files=*_tests.py
+testpaths=djmoney
+norecursedirs=testproject .svn _build tmp* dist *.egg-info
 DJANGO_SETTINGS_MODULE=settings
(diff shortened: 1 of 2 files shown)


benjaoming:
Merging this, even though there are 3 failures for django 1.8+ and python 3.3

@toudi would you be able to look at this once rebased into one of your PRs?

toudi:
I have seen the test results and some of the errors were fixed in my commits (i.e. the 2 vs 3 as of the queryset length)

I will rebase it once more and push so that the code can pick up the newest tests ;)

benjaoming:
Awesome, thanks so much :D

PR #164, opened 2016-01-08
Set latest Django versions for tox, simplify config for Travis CI

Stranger6667:


benjaoming:
Thanks for this nice simplification, looks like a `$` is missing in the Travis conf

Stranger6667:
I've squashed two last commits, now everything is fine. Also I'll add Python 3.5 support here

benjaoming:
:+1: for 3.5 support! I might have been in an outdated browser window with the previous comment.

PR #168, opened 2016-01-15
[WIP] Extend F expressions support

Stranger6667:
I want to add some checks. For such cases:
```python
F('money_field') * F('money_field')
```
We can't multiply money instances, but now it is possible to do that, because it is numeric fields.
Also I'll check F expressions in `.update()`


Stranger6667:
Sorry, wrong base branch

PR #169, opened 2016-01-15
Extend F expressions support. Closes #167

### LINKED ISSUE #167: Extend F expressions support
This code works:
```python
instance = ModelWithVanillaMoneyField.objects.create(money=Money(100, 'USD'))
instance.money = F('money') + Money(100, 'USD')
instance.save()
```
Same for subtraction.
Multiplication and division is not supported yet, so this code fails:
```python
instance = ModelWithVanillaMoneyField.objects.create(money=Money(100, 'USD'))
instance.money = F('money') * 2
instance.save()
```
With
```python
    def _check_currency(obj, field, amt):
        currency_field_name = get_currency_field_name(field.name)
>       if obj.__dict__[currency_field_name] != str(amt.currency):
E       AttributeError: 'int' object has no attribute 'currency'

djmoney/models/fields.py:208: AttributeError
```
I'll create new PR for this issue after #165 will be merged to master (I hope it will :))

Stranger6667:
I want to add some checks. For such cases:
```python
F('money_field') * F('money_field')
```
We can't multiply money instances, but now it is possible to do that, because it is numeric fields.
Also I'll check F expressions in `.update()`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: python
 python:
   - 3.5
-install: pip install tox codecov
+install: travis_retry pip install "virtualenv<14.0.0" tox codecov
 script: tox -e $TOX_ENV
 env:
   - TOX_ENV=django_master-py35
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ help:
 	@echo "coverage - check code coverage quickly with the default Python"
 	@echo "install - install the package to the active Python's site-packages"
 
-clean: clean-build clean-pyc clean-test
+clean: clean-test clean-build clean-pyc
 
 clean-build:
 	rm -fr build/
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -7,6 +7,7 @@ from decimal import ROUND_DOWN, Decimal
 from django import VERSION
 from django.conf import settings
 from django.db import models
+from django.db.models import F
 from django.db.models.signals import class_prepared
 from django.utils import translation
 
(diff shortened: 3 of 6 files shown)


Stranger6667:
Also this will not work:
```python
Money(100, 'USD') + F('money')
```

Stranger6667:
Please, check if I added correct cases for valid and invalid expressions.
And what we can do with
```python
Money(100, 'USD') + F('money')
```
?
May be some monkeypatching? py-moneyed doesn't know about Django's F objects and rises exception on that statement.

Stranger6667:
Also I'm not sure about invalid expressions for MoneyField, please advise

benjaoming:
> Money(100, 'USD') + F('money')

Is this supposed to be possible? The F query references the current state of a DB column, if we refer it like this, we can't check that the currency is the same. I mean, ideally it would generate this:

   money__amount=100 + money__amount WHERE (...) and money__currency='USD'

...but that's not possible.

So raising an exception is perhaps the best? I.e. "unsupported operation, cannot add known currency to unknown currency blah"?

Stranger6667:
Unfortunately 
```python
Money(100, 'USD') + F('money')
```

Will fail instantly with something like that:
```python
tests/test_models.py:42: in <module>
    class TestVanillaMoneyField:
tests/test_models.py:90: in TestVanillaMoneyField
    (Money(100, 'USD') + F('money'), Money(200, 'USD')),
/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/moneyed/classes.py:109: in __add__
    'Money and non-Money instance.')
E   TypeError: Cannot add or subtract a Money and non-Money instance.
```

So, it relies on `Money.__add__` implementation inside `py-moneyed`, but in fact these things should be equal:
```python
Money(100, 'USD') + F('money')
```

```python
F('money') + Money(100, 'USD')
```

I didn't think about the implementation on `django-money` side, but it may be tricky, I'll try to do that.

>  I mean, ideally it would generate this:
> money__amount=100 + money__amount WHERE (...) and money__currency='USD'
> ...but that's not possible.
> So raising an exception is perhaps the best? I.e. "unsupported operation, cannot add known currency to unknown currency blah"?

Such checks are implemented [here](https://github.com/django-money/django-money/pull/169/files#diff-a3ba76a61c67ec38bfe7ac546e2548e0R164)

About permitted operations

Preliminary I've decided to disallow following operations:
1. Any operations on money instances with different currencies. This one seems OK to me
2. Exponentiation
3. Add or subtract money with not-money
4. Multiplication, division, modulo with money instances on both sides of expression

In 2-4 I rely on `py-moneyed` `Money` implementation, where these actions are disallowed. But in Django we have decimal field with value, and these value can be used in all arithmetic operations. So, if someone wants to deal with decimal part of `MoneyField` these operations will fail, which may cause some inconvenience for users. And we have to decide here about expected behavior. May be the most expected behavior here is to allow users to deal with decimal part of `MoneyField` without restrictions on arithmetic operations. What do you think?

benjaoming:
I would think that if one does addition and subtraction with Money, then forcing both sides to be a Money data-type is fine. Source data of this operation *should* be Money, otherwise there is an implicit understanding that the non-Money side of the expression *is* Money..

As for multiplication and division, it should be possible to have one of the two sides be non-Money.

I see that the limitation is in py-moneyed, and I think the patch work should be done there.

> May be the most expected behavior here is to allow users to deal with decimal part of MoneyField without restrictions on arithmetic operations. What do you think?

If you do operations like `Money(money.amount * factor, money.currency)`, you have all the freedom.. that's a reasonable workaround. But I'm not sure if that's what you meant?

Stranger6667:
Yes, you are right :) I didn't think about it in this way. I'll try to implement some workaround for this py-moneyed limitation

Stranger6667:
Another option - use MoneyPatched class to adjust such limitation?
E.g. we rename MoneyPatched --> Money and expose it like part of public API:

```python
from djmoney import Money

Model.objects.update(model=Money(100, 'USD') + F('money')
```

And this `Money` class will implement all the behaviors related to Django (e.g. this F expression issue), localization and conversion (which it already does).

benjaoming:
Hey there :) Is this still WIP?

Stranger6667:
Hi! Yes, I plan to finish it this week :)

Stranger6667:
I think, that for compatibility reasons it will be better to consider this PR as completed.
Further changes (`MoneyPatched` class update) for F extensions support requires more work at this point, so I'll continue it in separate PR.
Thanks

benjaoming:
No worries! Thanks for all the work thus far! Please shout or open an issue if you want a new release.

PR #177, opened 2016-02-12
Unified logic and reduced code duplication

Stranger6667:
I found that some logic involved in expressions handling is duplicated


benjaoming:
Great refactor, very understandable. Perhaps soon we can pass the McCabe test :)

Issue #172, opened 2016-02-08
Updating the currency of an existing object fails


Hedde:
We are trying to update the currency of an existing object (django-money 0.7.6), this fails because of a specific check in the set method, what is the idea of this check?

    def __set__(self, obj, value):
        if isinstance(value, tuple):
            value = Money(amount=value[0], currency=value[1])
        if isinstance(value, Money):
            obj.__dict__[self.field.name] = value.amount
            # we have to determine whether to replace the currency.
            # i.e. if we do the following:
            # .objects.get_or_create(money_currency='EUR')
            # then the currency is already set up, before this code hits
            # __set__ of MoneyField. This is because the currency field
            # has less creation counter than money field.
            obj_curr = obj.__dict__[self.currency_field_name]
            val_curr = str(value.currency)
            def_curr = str(self.field.default_currency)
            if obj_curr != val_curr:
                # in other words, update the currency only if it wasn't
                # changed before.
                if obj_curr == def_curr:   # <=== why ????
                setattr(
                    obj, self.currency_field_name,
                    smart_unicode(value.currency))

Stranger6667:
@benjaoming can you advice on that? What suppose to be expected here?
If we want to change the money value of an existing object. IMO:
1. In all cases, when the currency is explicitly set - provided value should be used. No matter what kind of operation it is.
2. If there is no value for currency field, then default currency should be used

benjaoming:
I would agree to your list! And also that the above code seems like it was written in a rush, hard to understand what the intention of the check is.

Stranger6667:
@Hedde , could you check the issue with the latest version of `django-money`? It is not on PyPI yet, but this behavior was fixed few commits ago

andir:
I've tried with the pending 0.8 release branch from @Stranger6667 and with the current master. Both seem to revert to the previous currency once I do a full clean on my model.

My test case looks similar to
```python
obj = AccountFactory()
obj.amount # is the Moneyfield
# change from default currency (USD) to AUD
obj.amount.currency = moneyed.Currency(code='AUD')
obj.full_clean()
# afterwards this fails
assert obj.amount.currency.code == 'AUD'
# this succeeds...
assert obj.amount.currency.code == 'USD'
```

Stranger6667:
Thank you for the report!
Could you please tell which Django version did you use?

andir:
django 1.9

Stranger6667:
You can use the following code:
```
[...]

PR #178, opened 2016-02-12
Changed currency handling behaviour. Changed exception type

Stranger6667:
Changes relates to #172.
Also exception type was changed to be consistent with `DecimalField.to_python` behaviour

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -343,6 +343,8 @@ class MoneyField(models.DecimalField):
             value = value.amount
         if isinstance(value, tuple):
             value = value[0]
+        if isinstance(value, float):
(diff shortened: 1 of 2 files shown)


Stranger6667:
Oops, I'll fix that

benjaoming:
Nice, clean solution! Thanks! That seems to fix #172 -- maybe @Hedde can comment? I'm merging it now, but would be nice with a confirmation from the original reporter.

n4cr:
@Stranger6667 @benjaoming  This still has the issue in django admin. If I change the currency from anything else to default currency then it won't get updated. Mainly because of this check:

```
 if object_currency != value and (object_currency == default_currency or value != default_currency):
```

PR #182, opened 2016-02-24
Fix access to manager via model instances. Closes #86. Closes #87.

### LINKED ISSUE #86: Instances of models with MoneyFields have "objects" attribute

Stranger6667:
Managers should not be accessible via model instances.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -467,8 +467,10 @@ def patch_managers(sender, **kwargs):
     from .managers import money_manager
 
     if hasattr(sender._meta, 'has_money_field'):
-        for _id, name, manager in sender._meta.concrete_managers:
-            setattr(sender, name, money_manager(manager))
+        sender.copy_managers([
+            (_id, name, money_manager(manager))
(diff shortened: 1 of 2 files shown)


djmoney/models/fields.py
```diff
@@ -467,8 +467,10 @@ def patch_managers(sender, **kwargs):
     from .managers import money_manager
 
     if hasattr(sender._meta, 'has_money_field'):
-        for _id, name, manager in sender._meta.concrete_managers:
-            setattr(sender, name, money_manager(manager))
+        sender.copy_managers([
```
benjaoming: Yup, that's a very backwards-compatible function, found it in dj 1.4, too! :+1:

benjaoming:
Awesome! Thanks for the tests, too! I've made you repo admin btw, thanks for all the nice work! It wasn't to make you feel like I don't wanna review PRs, I'll happily keep doing that because it's always nice to share responsibility!

PR #185, opened 2016-03-01
Fixed m2m data loss for Django < 1.8. Closes #184

### LINKED ISSUE #184: Bug in deserialization
It is a regression introduced in e9d02c546ffeb726ea43084ed4a8153d40c2a6f0.
Before that commit the `obj['fields']` included all fields and m2m relations like `user_permissions` and `groups` on the `User` model.
The commit removes M2M fields from deserialized data, which leads to data loss.

I'll submit PR with tests and fix.

Stranger6667:
For me this bug cause errors in `generate_patch_html` inside django-reversion. I'll add some refactoring and tests for serializer in separate PR.

Best regards, Dmitry

### DIFF
--- a/djmoney/_compat.py
+++ b/djmoney/_compat.py
@@ -56,3 +56,14 @@ def set_expression_rhs(expr, value):
         expr.children[1] = value
     else:
         expr.rhs.value = value
+
+
+def get_fields(model):
+    """
+    Returns a set of fields associated to the model.
+    """
+    opts = model._meta
+    if VERSION < (1, 8):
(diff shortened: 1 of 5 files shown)


benjaoming:
Grr, sad that reversion causes more headaches. In my experience, using it was never a benefit.. in the end I always ended up writing a custom-made revision system... django-reversion was just creating user-unfriendly garbage in the DB and doesn't work once stuff is migrated away from the originally serialized schema.

Anyways! If this work, I don't see any problems since reversion is widely used.

PR #187, opened 2016-03-03
Fix fields caching

Stranger6667:
In Django fields comparison is based on `creation_counter` attribute, so CurrencyField and MoneyField have the same creation_counter value. This leads to problems with caching in model options


Stranger6667:
It is deprecated API, but many libraries still use it. E.g. django-filters

PR #190, opened 2016-03-13
Release 0.8. Closes #189

Stranger6667:
This PR adds rst version of changelog and bumps version of django-money

PR #191, opened 2016-03-13
Added __version__ attribute. Closes #89.

### LINKED ISSUE #89: Add __version__ attribute
It would be nice to add a `__version__` attribute (as specified in http://legacy.python.org/dev/peps/pep-0396/) to the djmoney module, so it's possible to check which version is installed.

Thanks!

Stranger6667:
For now I don't see anything against this attribute. @benjaoming , what do you think?

You wrote in #89 :
> Once we have fixed the \__init\__.py file to stop importing and monkey patching over in #127 we could add this.

Why does this matter?

Best regards, Dmitry

P.S. If everything is OK, I'll update `__version__` in #190

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,19 @@
 # -*- coding: utf-8 -*-
+import os
+import re
 import sys
 
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
+
+
+def get_version(package):
+    """
+    Return package version as listed in `__version__` in `__init__.py`.
+    """
+    with open(os.path.join(package, '__init__.py')) as fd:
+        init = fd.read()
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,10 +1,21 @@
...
 
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
 
 
+def get_version(package):
+    """
+    Return package version as listed in `__version__` in `__init__.py`.
+    """
+    with open(os.path.join(package, '__init__.py')) as fd:
```
benjaoming: This looks way too hackish IMO :)

benjaoming:
The real fix is still to stop having code in `__init__.py`, it's uncool. It should be importable without side effects. This would also make it straight forward for usage in `setup.py`.

PR #192, opened 2016-03-19
Let MoneyPatched to be deconstructed in migrations

AlexRiina:
In some situations, `Money` objects will need to be serialized in  migrations. My particular case was a field like
```
class MyModel(models.Model):
    amount = MoneyField(null=True, max_digits=9, decimal_places=2, validators=[MinValueValidator(Money(0, 'USD'))])
```

which triggered
```
ValueError: Cannot serialize: 0 USD
There are some values Django cannot serialize into migration files.
For more, see https://docs.djangoproject.com/en/1.9/topics/migrations/#migration-serializing
```

But with `@destructible` wrapping the class, will work on django > 1.7.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,6 +81,21 @@ Searching for models with money fields:
         BankAccount.objects.filter(balance__gt=Money(1, USD))
         # Returns the "account" object
 
+Special note on serialized arguments: if your model definition 
+requires serializing an instance of ``Money``, you can use ``MoneyPatched``
+instead.
+
+.. code:: python
+
+        from django.core.validators import MinValueValidator
+        from django.db import models
+        from djmoney.models.fields import MoneyField, MoneyPatched
+
+        class BankAccount(models.Model):
+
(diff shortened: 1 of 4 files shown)


AlexRiina:
fwiw, the work around I've been using is just wrapping the validators in a simple function in my models file.

Stranger6667:
Thank you for the PR! I'll take a look tomorrow, but at first glance it seems pretty good :+1:

README.rst
```diff
@@ -81,6 +81,23 @@ Searching for models with money fields:
...
         # Returns the "account" object
 
+Special note on serialized arguments: if your model definition 
+requires serializing an instance of ``Money``, you can use ``MoneyPatched``
+instead.
+
+.. code:: python
+
+        from django.core.validators import MinValueValidator
+        from djmoney.money
```
Stranger6667: This line is incomplete

AlexRiina:
Woops. I've removed that extra import statement from the `README`

Stranger6667:
Thank you! :+1:

PR #194, opened 2016-05-09
Patch potential problem to avoid overriding other managers in same model

inureyes:
Current code has a potential problem when model uses two or more managers. This code overwrites every manager to ``MoneyManager``, which causes critical errors during migration.

e.g. If model has ``CurrentSiteManager`` as a field,
```
on_site = CurrentSiteManager()
```

``make migrations`` command generates the migration code to alter ``on_site`` manager from ``CurrentSiteManager`` to ``MoneyManager``.

This patch only overwrites model's ``objects``. Code will work as described in README.md. (default is ``objects``, and should use ``money_manager`` for other fields)

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -463,7 +463,7 @@ def patch_managers(sender, **kwargs):
     if hasattr(sender._meta, 'has_money_field'):
         sender.copy_managers([
             (_id, name, money_manager(manager))
-            for _id, name, manager in sender._meta.concrete_managers
+            for _id, name, manager in sender._meta.concrete_managers if name == 'objects'
         ])
 
 


benjaoming:
Thanks for fixing this!

Let's say this is for a 0.9 release because it's potentially as breaking as it's fixing... Could you add a note here that only `objects` is now patched by default and a note on how to patch managers with other names?

https://github.com/django-money/django-money/blob/master/CHANGES.rst

inureyes:
Here it is:

Note about automatic model manager patches
===================================
In 0.8, Django-money automatically patches every model managers with MoneyManager. This causes migration problems if two or more managers are used in the same model. 

As a side effect, other managers are also finally wrapped with `MoneyManager`. This effect leads Django migration to point to fields with other managers to `MoneyManager`, and raises ``ValueError`` (`MoneyManager` only exists as a return of money_manager, not a class-form. However migration procedure tries to find `MoneyManager` for other managers.)

From 0.9, Django-money only patches `objects` with `MoneyManager` by default (as documented). To patch other managers (e.g. custom managers), patch them by wrapping with `money_manager`.

```!python
from djmoney.models.managers import money_manager
class BankAccount(models.Model):
    balance = MoneyField(max_digits=10, decimal_places=2, default_currency='USD')
    accounts = money_manager(MyCustomManager())
```

benjaoming:
@inureyes sorry about not getting back on this, I must have switched off notifications during that period.

I'm going to merge your PR after rebasing it on the current master. I'll make sure to include the above note, thanks for writing it!!

Issue #199, opened 2016-06-03
Auto conversion of currency is a really bad default, plus you can't change it


spookylukey:
I'm talking about the behaviour added here:

https://github.com/django-money/django-money/commit/edfc25a5aa46a605f27d15b891bfae28b74cb5db

which I discovered when upgrading from 0.7.2

Can I persuade you this is a really bad idea? I think it is a very bad idea for any auto-conversion to be built into the money class at all, but having it on by default, and without a way to turn it off is really really bad, because:

1) currency conversion is not a lossless process. You simply can't "compare" two objects in different currencies, nor add them, or subtract them. You can't "compare" potatoes to tomatoes simply because you can use one to buy the other, and for the same reason you can't do the same with currencies, because the price changes all the time. In fact, even at a single point in time there is not one answer - there are many answers, different bodies will trade the commodity at different prices. Exchange rate indexes are just averages.

In addition to changing market exchange rates, there could be any number of additional things you want to do when changing currency. For example, because of rate changes, it's entirely possible that what someone was supposed to pay is no longer the 'right' price. You need to have a strategy for coping with that (e.g. rounding a bit) and silent currency conversions add confusion not help.

Instead, if you want to compare currencies in any way, you should be explicitly doing a conversion. "Explicit is better than implicit" etc. As Python matures, it always gets closer to that ideal e.g. byte strings and unicode cannot be compared in Python 3.

In this way, if you accidentally compare or add two money objects of different currencies, you made a mistake as a developer, and should be informed with a loud exception.

2) You are changing the behaviour of another library.

With plain py-moneyed, as above, you get the desired behaviour:

    >>> Money(10, GBP) > Money(10, USD)
    TypeError: Cannot compare Money with different currencies.

    >>> Money(10, GBP) + Money(10, USD)
    TypeError: Cannot add or subtract two Money instances with different currencies.

It's extremely surprising when django-money doesn't behave the same way, when it is supposed to be built on top of it, and is advertised as such. This exception is extremely important part of how py-moneyed works and what I depend on for ensuring correctness of my app. I happen to use djmoney_rates too, but only ever with explicit conversions.

This is also a really big change from 0.7.2, when it did behave the same as py-moneyed.

3) currency conversion introduces rounding that you didn't ask for, even if the currencies are the same - see https://github.com/evonove/django-money-rates/blob/master/djmoney_rates/utils.py#L55

[...]

PR #202, opened 2016-06-08
Change default autoconvert behaviour

spookylukey:
See issue #199

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,9 @@
 Changes in 0.8.1
 ----------------
 - Added Django 1.10 support `198 <https://github.com/django-money/django-money/issues/198>`_ (`Dmitry Dygalo <https://github.com/Stranger6667>`_)
+- Changed auto conversion of currencies using djmoney_rates (added in 0.7.3) to
+  be off by default. You must now add ``DJMONEY_AUTO_CONVERT_MONEY = True`` in
+  your ``settings.py`` if you want this feature.
 
 Changes in 0.8
 --------------
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -34,13 +34,15 @@ from ..utils import get_currency_field_name, prepare_expression
 if 'djmoney_rates' in settings.INSTALLED_APPS:
     try:
         from djmoney_rates.utils import convert_money
-        AUTO_CONVERT_MONEY = True
+        RATES_INSTALLED = True
     except ImportError:
         # NOTE. djmoney_rates doesn't support Django 1.9+
-        AUTO_CONVERT_MONEY = False
+        RATES_INSTALLED = False
 else:
-    AUTO_CONVERT_MONEY = False
+    RATES_INSTALLED = False
 
+
+AUTO_CONVERT_MONEY = RATES_INSTALLED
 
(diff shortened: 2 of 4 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,9 @@
 Changes in 0.8.1
 ----------------
 - Added Django 1.10 support `198 <https://github.com/django-money/django-money/issues/198>`_ (`Dmitry Dygalo <https://github.com/Stranger6667>`_)
+- Changed auto conversion of currencies using djmoney_rates (added in 0.7.3) to
```
benjaoming: With this change in mind, would you mind bumping the version for all the above unreleased changes to 0.9 (changing the subheading)

The version itself is bumped in setup.py

benjaoming:
Looks ripe! Great stuff @spookylukey !

CHANGES.rst
benjaoming: *doh* I already did the version bump in #201 

Sorry if you're getting conflicts now........

djmoney/models/fields.py
```diff
@@ -34,12 +34,17 @@
...
     except ImportError:
         # NOTE. djmoney_rates doesn't support Django 1.9+
-        AUTO_CONVERT_MONEY = False
+        RATES_INSTALLED = False
 else:
-    AUTO_CONVERT_MONEY = False
+    RATES_INSTALLED = False
+
+
+if getattr(settings, 'DJMONEY_AUTO_CONVERT_MONEY', False):
```
Stranger6667: This line is duplicated (at line 79), so another constant with ``getattr(settings, 'DJMONEY_AUTO_CONVERT_MONEY', False)`` will be useful

djmoney/models/fields.py
```diff
@@ -71,7 +76,7 @@ def _convert_to_local_currency(self, other):
         """
         Converts other Money instances to the local currency
         """
-        if AUTO_CONVERT_MONEY:
+        if getattr(settings, 'DJMONEY_AUTO_CONVERT_MONEY', False):
```
benjaoming: Something like `djmoney.settings.AUTO_CONVER_MONEY_ENABLED` would be cool?

djmoney/models/fields.py
spookylukey: I did it that way at first, but then you end up with a module level constant, and can't change it (easily) in the tests, at least, not with the normal 'settings' fixture provided by pytest-django - https://pytest-django.readthedocs.io/en/latest/helpers.html#settings - you would have to do some other kind of monkey patching of that constant. I thought this duplication was the lesser of two evils. What do you think?

CHANGES.rst
spookylukey: I didn't do anything on this yet - are you saying we don't need a version bump? The latest released version is 0.8 right?

Also are you OK with the code duplication above for the reasons given?

benjaoming: It's because I've merged another PR meanwhile that bumped the version, so you'll need to fix a conflict here...

spookylukey: OK, fixed this now by merging from master.

benjaoming:
Rebased in #207

PR #204, opened 2016-07-02
Added ``in`` lookup support

Stranger6667:


Stranger6667:
@benjaoming 
Please, take a look if you'll have time:) It should be a useful feature

tests/test_models.py
```diff
@@ -128,15 +140,8 @@ def test_rounding(self):
             (Q(amount1=Money(1, 'USD')) | Q(amount1__gte=Money(4, 'USD'), amount2=Money(0, 'GHS')), 2),
         )
     )
+    @pytest.mark.usefixtures('objects_setup')
```
benjaoming: That's pretty cool :)

benjaoming:
Thinking:

1. It's tested
1. It's a new feature
1. It's useful!

I think it should be merged, but could you wait for #207 ? That way, we can have it cleanly integrated with the new release notes etc.

Been having quite some problems rebasing it and waiting for Travis.. so far, rebasing and merging a simple PR has prolonged for 24+ hours, I think I've done like 3-4 force-pushes already :/

tests/test_models.py
Stranger6667: I think I'll refactor whole querying tests with single objects setup. But it can wait, and I hope, that such approach will simplify our tests

Stranger6667:
Sure we can wait for #207. (Un)fortunately we have a lot of builds. On the other side, we have support for Django 1.4-1.7 without big complications. What do you think, should we drop support for old Django versions?

benjaoming:
@Stranger6667 I think we should keep it as long as it comes almost for free, but if we encounter backwards incompatibilities when implementing new features, we should drop it. People on old django versions can also use old django-money versions :)

PR #207, opened 2016-07-05
Spookylukey change default autoconvert behaviour

benjaoming:
Supersedes #202

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -11,10 +11,9 @@ from django.db.models import F
 from django.db.models.signals import class_prepared
 from django.utils import translation
 
+from djmoney import forms
 from moneyed import Currency, Money
 from moneyed.localization import _FORMATTER, format_money
-
-from djmoney import forms
 
 from .._compat import (
     BaseExpression,
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -5,7 +5,7 @@ import inspect
 from decimal import ROUND_DOWN, Decimal
 
 from django.conf import settings
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, ImproperlyConfigured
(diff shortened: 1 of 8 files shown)


djmoney/models/fields.py
```diff
@@ -34,12 +34,17 @@
 if 'djmoney_rates' in settings.INSTALLED_APPS:
     try:
         from djmoney_rates.utils import convert_money
```
Stranger6667: IMO it is a bad thing - to import models on module level. In some cases it leads to "app registry is not ready" errors.

tests/test_models.py
```diff
@@ -351,7 +347,7 @@ def test_patching(self):
 
 
 rates_is_available = pytest.mark.skipif(
-    not AUTO_CONVERT_MONEY,
+    not RATES_INSTALLED,
```
Stranger6667: Seems it is not imported

djmoney/models/fields.py
benjaoming: These things should be handled on a conventional level IMO.

A convention to *never* download anything in `djmoney/__init__.py` is very reasonable! We already have an issue open on this.

I've opened up #208 to explicitly deal with this.

tests/test_models.py
benjaoming: Yeah, absolutely right.. something seems to have been renamed... added in 9c88515

Stranger6667:
I'll take care about this PR

benjaoming:
You might have to re-open it. I can't make Travis build the correct commit.. the failing lint test isn't reproducable locally with the same version of isort, so I suspect Travis for building the wrong commit dbceff3.

benjaoming:
re-open = closing this and the creating a brand new PR

benjaoming:
Merged in #210

PR #210, opened 2016-07-09
Autoconvert

Stranger6667:


Stranger6667:
@benjaoming 
We had a limit concurrent jobs enabled on Travis. I disable it and now builds should be faster

PR #214, opened 2016-07-29
Fixed ``get_or_create`` crash, when ``defaults`` is passed. Closes #213.

### LINKED ISSUE #213: crasher in _expand_money_kwargs with `get_or_create(defaults=...)`
When using the `defaults` kwarg to `get_or_create`, I get a crashed with django-money:

Relevant bit of stacktrace (the first line is called a model `get_or_create(..., defaults={...})` though that isn't obvious.

      File "/home/luke/devel/clients/wolfandbadger.com/src/python/dashboards/designer_dashboard/designer/tests/utils.py", line 32, in setup_online_subscriptions
        percentage=0.40,
      File "/home/luke/.virtualenvs/wolfandbadger/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
        return getattr(self.get_queryset(), name)(*args, **kwargs)
      File "/home/luke/.virtualenvs/wolfandbadger/src/django-money/djmoney/models/managers.py", line 161, in wrapper
        args, kwargs = _expand_money_kwargs(model, args, kwargs)
      File "/home/luke/.virtualenvs/wolfandbadger/src/django-money/djmoney/models/managers.py", line 134, in _expand_money_kwargs
        elif isinstance(_get_field(model, name), MoneyField):
      File "/home/luke/.virtualenvs/wolfandbadger/src/django-money/djmoney/models/managers.py", line 63, in _get_field
        field = qs.setup_joins(parts, opts, alias)[0]
      File "/home/luke/.virtualenvs/wolfandbadger/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1472, in setup_joins
        names, opts, allow_many, fail_on_missing=True)
      File "/home/luke/.virtualenvs/wolfandbadger/lib/python2.7/site-packages/django/db/models/sql/query.py", line 1397, in names_to_path
        "Choices are: %s" % (name, ", ".join(available)))
    FieldError: Cannot resolve keyword 'defaults' into field. Choices are: ... [field names snipped]

It seems clear that this: https://github.com/django-money/django-money/blob/master/djmoney/models/managers.py#L134

is not taking into account the fact that `defaults` can be passed to `get_or_create` - https://docs.djangoproject.com/en/dev/ref/models/querysets/#get-or-create

Stranger6667:


benjaoming:
Great work @Stranger6667 :)

@spookylukey does this fix the issue as you intended?

Stranger6667:
Hi @benjaoming ! Thank you :)
Personally I think that it is not enough, because it will not work for custom methods on managers. E.g:

```
class MoneyManager(models.Manager):

    @understands_money
    def super_method(self, argument=None, **kwargs):
        return self.filter(**kwargs)
```

If #215 will be fixed, then understands_money will work only on methods without arguments, which should not be passed to queryset method. As ``defaults`` argument for ``get_or_create``
So, I'll refactor this code in the future. Probably I'll try introspect custom method

spookylukey:
Thanks - I haven't tried this yet, but it looks good to me.

@Stranger6667 - I would advise against introspection - it is generally very fragile, because if you change to kwargs at any point (for example many decorators do this), then it won't work.

It might make sense to split out `understands_money` into two versions, one for public use, which is clearly documented in terms of the exact signature it expects, and one for internal use.

Stranger6667:
Aha, I got your point. May be it is suitable only for original django queryset's methods

Stranger6667:
Hm, in Django 1.4 ``get_or_create`` has no ``defaults`` argument in signature, but it has ``defaults = kwargs.pop('defaults', {})``
So, this approach seems to be ok, introspection was a bad idea :(

benjaoming:
@reinbach we have another contestant for a patch release IMO :)

Stranger6667:
Also, I want to ask about permission to push new releases to PyPI. 
Is it possible for me and @benjaoming ?

benjaoming:
...let me add that I really enjoy @reinbach's taking responsibility for the releases, and I hope he'll continue... the more the merrier!! but definitely not a discouragement to pass around some permissions anyways just in case :)

reinbach:
Release 0.9 has been made.

Absolutely you guys can have permissions to push new releases, just let me know what your usernames are on pypi.

benjaoming:
"benjaoming" is mine :)

reinbach:
You've been added :)

Stranger6667:
And mine is "Stranger6667" :)

Stranger6667:
b34573cca3061dcfca620c676e95271f5fb84328
I screwed with packaging :( Sorry for this :(

PR #216, opened 2016-07-30
Fixed ``understands_money`` behaviour. Closes #215

### LINKED ISSUE #215: Bug in ``understands_money`` function
As the doc states this functions can be used as a decorator, but it is not possible because ``understands_money`` (misspelled in documentation as ``understand_money``) have 2 arguments

Stranger6667:


Stranger6667:
@benjaoming 
What do you think about copied code from ``six`` at ``_compat`` module?
I just didnt want to have ``six`` as explicit dependency

benjaoming:
@Stranger6667 I guess Django also ships with a copy? (but I don't know if it's safe to assume the version of it)

`six` is made for bundling so sounds cool :)

Stranger6667:
Yes, but only ``six`` 1.7.0+ has ``wraps``, which is required to get the model, so it is ported from ``six`` 1.7.0

Issue #225, opened 2016-09-19
Comparison with different currencies


ivirabyan:
Can you explain why MoneyPatched class doesn't allow equality comparison of two values if they are of different currency? It seems pretty ok to know if `Money(100, 'USD') == Money(100, 'EUR')`.
When you try to do this with MoneyPatched class, you get an error "Cannot add or subtract two Money instances with different currencies". Seems like the message was copypasted from other method, and the error is non relevant here.

benjaoming:
Spot on :)

https://github.com/django-money/django-money/blob/master/djmoney/models/fields.py#L106

ivirabyan:
Yeah, I saw that line, and it makes me think it was incorrectly copypasted from some other place

benjaoming:
It looks like it indeed :)

But the intention seems okay, I mean not allowing comparison of different currencies. It's better to just fail hard.

ivirabyan:
Comparison in general is really somewhat difficult, but simple equal/non equal is perfectly fine. At least original Money type does that. The purpose of this project is to bring Money type to Django as a model field. Why change it's behaviour?

benjaoming:
> The purpose of this project is to bring Money type to Django as a model field. Why change it's behaviour?

Well, even though it seems accidental, the behavior is already implemented :)

But would be interesting to hear why you would like it to return `False` instead?

ivirabyan:
I simply want to know if the field value changed. Suppose I stored initial field value upon model instantiation, and on save I want to check if it changed.
```
if old_price != self.price:
   # do something
```
That's pretty normal thing to do. At least we must have on option for such behaviour. Right now I had to subclass three classes to just revert to the original behaviour: MoneyField, MoneyFieldProxy, MoneyPatched. It's very overwhelming work to just change one line of code, so I'd like it to be in an original code, at least as an option.

benjaoming:
I'm convinced, thanks :)

benjaoming:
@Stranger6667 any comments? I'm just PR'ing it now...

benjaoming:
PR updated w/ new behavior + release note.

ivirabyan:
Thanks!

Stranger6667:
I agree with @ivirabyan

vvkuznetsov:
> Right now I had to subclass three classes to just revert to the original behaviour: MoneyField, MoneyFieldProxy, MoneyPatched.

As a man who did it, I agree. Thanks for fast change.

[...]

Issue #220, opened 2016-09-13
Thousand separator added twice in some views


benjaoming:
Original post by @k8n

> Not sure if this is related or merits a separate issue ticket.
> I am seeing things like "9,,999.99" in admin changelists (note the double comma).
> I have this in settings.
> 
> ```
> USE_L10N = True
> USE_THOUSAND_SEPARATOR = True
> ```
> 
> Admin treats MoneyField field as a DecimalField and applies L10N formatting on top of the already-formatted value. Essentially the following happens:
> 
> ```
> from django.utils import formats
> ...
> formats.number_format(field, decimal_places)
> ...
> ```
> 
> See https://github.com/django/django/blob/master/django/contrib/admin/utils.py#L413.

To make it easier to solve this, it would be nice to know exactly which places you see the double comma @k8n?

https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-USE_THOUSAND_SEPARATOR

k8n:
@benjaoming thanks.

Workarounds for this issue are plenty, of course. It's just that a double-comma doesn't look right as the default behaviour.

The issue is apparent wherever django admin renders a naked `MoneyField` when both `USE_THOUSAND_SEPARATOR` and `USE_L10N` are `True`. That happens, for example, when viewing a model's change list in admin.

admin's `display_for_field()` method treats `MoneyField` as a `DecimalField` and calls `number_format()` on it. Both `number_format()` and MoneyField's `__unicode__()` do localization formatting.

I couldn't yet think of a simple / clever solution that would be backward-compatible.

I looked at the use of the `use_l10n` property on the `MoneyPatched` class in hopes to control this localization behaviour. It seems that in this context, the value of `use_l10n` makes no difference to the output.

Trouble is too, even if it worked, `use_l10n` cannot be conveniently set. It may be nice to have `MoneyField`'s constructor pass this property to `MoneyPatched` somehow.

Here is more of an example of this behaviour in django's shell:

```
from django.conf import settings
from django.db import models
from djmoney.models.fields import MoneyField
from djmoney.models.fields import Money
from django.utils import formats
from django.contrib.admin.utils import display_for_value

settings.USE_L10N = True
settings.USE_THOUSAND_SEPARATOR = True

class A(models.Model):
    m = MoneyField()
    class Meta:
        app_label = "test"

a = A()
a.m = Money(1234) 
formats.number_format(a.m, 2)  # returns u'1,,234.00'

a.m.use_l10n = False
[...]

PR #226, opened 2016-09-19
Correct wrong exception message when comparing different currencies 

benjaoming:
#225

Shuffling import because of isort

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@ Changes in 0.9.2
 ----------------
 
 - Fixed ``understands_money`` behaviour. Now it can be used as a decorator `#215`_ (`Stranger6667`_)
+- Do not fail comparisons because of different currency. Just return ``False`` `#225`_ (`benjaoming` and `ivirabyan`)
 
 Changes in 0.9.1
 ----------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-Changes in 0.9.2
+Changes in 0.10
 ----------------
 
 - Fixed ``understands_money`` behaviour. Now it can be used as a decorator `#215`_ (`Stranger6667`_)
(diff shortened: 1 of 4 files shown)


benjaoming:
Maybe we should bump to 0.10 instead since it's breaking. That way, we can also introduce a fix for #220

Stranger6667:
Hm, you're right. 0.10 is a way to go

Stranger6667:
But I think, that it would be better to include some more changes to new release :) I'll submit some in a next few days

benjaoming:
https://en.wikipedia.org/wiki/Release_early,_release_often :)

benjaoming:
Don't stress adding more changes in order to release, it's totally okay to do a release with just one change IMO :)

Stranger6667:
okey:)

CHANGES.rst
```diff
@@ -1,7 +1,8 @@
-Changes in 0.9.2
+Changes in 0.10
 ----------------
 
 - Fixed ``understands_money`` behaviour. Now it can be used as a decorator `#215`_ (`Stranger6667`_)
+- Do not fail comparisons because of different currency. Just return ``False`` `#225`_ (`benjaoming` and `ivirabyan`)
```
Stranger6667: To make links work properly it should be:
```
`benjaoming`_ and `ivirabyan`_
```

CHANGES.rst
```diff
@@ -217,4 +218,4 @@ Changes in 0.3
 .. _toudi: https://github.com/toudi
 .. _tsouvarev: https://github.com/tsouvarev
 .. _w00kie: https://github.com/w00kie
-.. _willhcr: https://github.com/willhcr
\ No newline at end of file
+.. _willhcr: https://github.com/willhcr
```
Stranger6667: Please, add

```
.. ivirabyan: https://github.com/ivirabyan
```

benjaoming:
@Stranger6667 - thanks for the comment, addressed and pushed new changes + rebased to master.

Stranger6667:
I'll fix sorting :) Thank you

PR #230, opened 2016-09-23
Update tests. Closes #229

### LINKED ISSUE #229: Fix Django master builds

Stranger6667:


tests/test_form.py
```diff
@@ -106,4 +106,8 @@ def test_default_currency():
     """
     instance = NullMoneyFieldModel.objects.create()
     form = NullableModelForm(instance=instance)
-    assert '<option value="USD" selected="selected">US Dollar</option>' in form.as_p()
+    if VERSION[:3] > (1, 10, 1):
+        expected = '<option value="USD" selected>US Dollar</option>'
+    else:
+        expected = '<option value="USD" selected="selected">US Dollar</option>'
```
benjaoming: Rant: Perhaps this is an example of expecting something that's a bit to precise -- not saying to change it, but what's wrong with this test is that it's prohibiting that things _change_, not that they're actually broken. A better test would just ensure that it's a string containing `<option` and `US Dollar`.

Issue #217, opened 2016-07-31
Admin viewer for development


Stranger6667:
It can be very handy to have an ability to start django app with admin with single command in any tox environment

Stranger6667:
It seems like a bit out of scope of the project

PR #231, opened 2016-09-23
Refactoring of a way of running tests

Stranger6667:
Hi!
What do you think about such approach?
It speeds up tests in about 20% on my machine, also there is no need to specify tests dependencies in 2 places - now only in tox.ini.
Another thing - it allows us to implement admin viewer for development without adding `tests` module to packages in `setup.py` (#217)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@ envlist =
     django{16,15}-py{33,32,27,26,py,py3}
     django14-py{27,26,py}
     lint
+skipsdist = true
 
 [testenv]
 deps =
@@ -17,14 +18,14 @@ deps =
     django19: {[django]1.9.x}
     django110: {[django]1.10.x}
     django_master: {[django]master}
+    py26,py27,py32,pypy: mock
     py32: coverage==3.7.1
     pytest<3.0.0
     pytest-django<3.0.0
     pytest-cov
     django-money-rates
-commands = make test
+commands = {posargs:py.test --ds=tests.settings --cov=./djmoney tests}
 usedevelop = true
-whitelist_externals = make
 
 [testenv:lint]
 deps =


README.rst
```diff
@@ -247,27 +247,20 @@ Formatting the number with currency:
...
 
 .. code:: bash
 
     tox
 
-Testing the application in the current environment python:
-
-.. code:: bash
-
-    make test
```
benjaoming: Not sure, why remove the test target? It's very convenient to be able to run tests quickly without invoking tox...?

tox.ini
```diff
@@ -17,14 +18,14 @@ deps =
     django19: {[django]1.9.x}
     django110: {[django]1.10.x}
     django_master: {[django]master}
+    py26,py27,py32: mock
```
benjaoming: where's mock being used?

tox.ini
```diff
@@ -17,14 +18,14 @@ deps =
...
     py32: coverage==3.7.1
     pytest<3.0.0
     pytest-django<3.0.0
     pytest-cov
     django-money-rates
-commands = make test
+commands = py.test --ds=tests.settings --cov=./djmoney tests
```
benjaoming: Are you saying that invoking `py.test` directly from tox is 20% faster than the Make -> setup.py -> py.test chain? :O

Makefile
```diff
@@ -27,13 +27,5 @@ clean-test:
 	rm -f .coverage
 	rm -fr htmlcov/
 
-test:
-	python setup.py test --pytest-args="--cov=./djmoney tests"
```
benjaoming: Make targets are a nice way of consolidating the entrypoints to a code base. Either as a means of quickly running commands or as a file that you read straight after the README to see how things are structured :)

tox.ini
Stranger6667: The only place is here - https://github.com/django-money/django-money/blob/master/tests/conftest.py#L25

tox.ini
Stranger6667: Yes, because here setuptools integration is not used at all + here I added `skipsdist = true`, which also speeds things a bit

Makefile
Stranger6667: You're right, I'll probably start only with tox.ini update :) Other things stay be untouched - I don't know how to solve duplicated dependencies problem yet, but anyway it will not make things worse - tox will have a complete list of dependencies without possible conflicts with setup.py.

Stranger6667:
@benjaoming 
Updated :)

So, current changes will simplify #217 a lot.

tox.ini
benjaoming: Ah okay! But somehow this test wasn't broken? Is it that it breaks when you have djmoney_rates installed?

tox.ini
benjaoming: Oh nice, thanks for the tip!! I guess it's only important if one has non-trivial functions in `setup.py`

benjaoming:
Nice work @Stranger6667 -- I look forward to seeing how that works! I guess it's something that would be great to a lot of reusable apps!

tox.ini
Stranger6667: Because setup.py installs ``mock`` package during ``python setup.py test``. 
It should not interfere with ``djmoney_rates``

PR #234, opened 2016-09-29
[WIP] Documentation

Stranger6667:
@benjaoming Hi!
I want to write the docs in this PR, so it is in progress now. I have a couple of questions to discuss.

### Author & copyright

Now I wrote this:

```python
copyright = u'2016, Jacob Hansson'
author = u'Jacob Hansson'
```

What do you think about?
Would "2011-2016, Jacob Hansson and contributors" be better?

### Structure

I propose to leave basic part + installation in `README.rst` and restructure the rest in something like this:

- **Usage**
    - **Model fields**. Information about how to define `MoneyField`, what is `MoneyField`, note about `CurrencyField`.
    - **Querying**. How to do queries. What is supported and what is not. Examples, `F` queries, `__in` lookups and limitations.
    - **Form fields**. How to use ``djmoney.forms.MoneyField`` and related widget. Here, IMO, we should consider some serious refactoring of our form field before writing the docs.
    - **Admin integration**. Probably, first of all we should fix our issues about admin behaviour.
    - **Template tags**. How to use `money_localize`.
    - **Django REST Framework support**. 
    - **Exchange rates**.
    - **Currencies**. How to add a new currency
- **Advanced topics**
    - **Managers**. How works our automatic patching, what to do with custom managers, how to define custom `QuerySet` methods.
    - **Serialization**. Some notes about `dumpdata` & `loaddata` behaviour.
    - **Migrations**. Just a few notes about migrations, serialization and `South` support.
- **Contribution**. How to contribute & test
- **Release Notes**.
    - Changelog entries
    - Known issues. Some extras, e.g. about Django 1.10.1 bug.

What do you think about such structure?


benjaoming:
> What do you think about?
> Would "2011-2016, Jacob Hansson and contributors" be better?

Yes, definitely.. especially since you're writing it in a totally new documentation where all the contents as of now are in fact `(c) Stranger6667` :)

Suggestions

 - I would put "known issues" in the release notes. Release Notes should be top-level.
 - Sort all sections into `Usage` and `Advanced topics` to make the top-level unreadable :)
 - Maybe don't have `extras` but create a clear topic for everything?

Stranger6667:
You mean something like this?

- **Usage**
    - **Model fields**. Information about how to define `MoneyField`, what is `MoneyField`, note about `CurrencyField`.
    - **Querying**. How to do queries. What is supported and what is not. Examples, `F` queries, `__in` lookups and limitations.
    - **Form fields**. How to use ``djmoney.forms.MoneyField`` and related widget. Here, IMO, we should consider some serious refactoring of our form field before writing the docs.
    - **Admin integration**. Probably, first of all we should fix our issues about admin behaviour.
    - **Template tags**. How to use `money_localize`.
    - **Django REST Framework support**. 
    - **Exchange rates**.
    - **Currencies**. How to add a new currency
- **Advanced topics**
    - **Managers**. How works our automatic patching, what to do with custom managers, how to define custom `QuerySet` methods.
    - **Serialization**. Some notes about `dumpdata` & `loaddata` behaviour.
    - **Migrations**. Just a few notes about migrations, serialization and `South` support.
- **Contribution**. How to contribute & test
- **Release Notes**.
    - Changelog entries
    - Known issues. Some extras, e.g. about Django 1.10.1 bug.

benjaoming:
That looks great! Some extra suggestions:

 - I would put a "know issues" headline for every release (because they change). Well, we don't have any back-log... but for the future, it's probably nicer to maintain lists separately for each release?
 - REST framework in "Advanced" - just to have a more equal number of subjects in each :)

Stranger6667:
Thanks :) So, probably exchange rates also should be in "Advanced"

I agree about known issues - if there is something to mention, then it should be mentioned in the release notes (If I understand you correctly)

benjaoming:
@Stranger6667 - what would you say to merging this and getting it on RTD and then finalizing it later? :)

Stranger6667:
agree :)

PR #238, opened 2016-11-15
Added Django REST Framework usage docs

mikhuang:
I'm not 100% sure this is how django-money is supposed to work with DRF but it works for me.


Stranger6667:
For Django 1.7+ it works automatically if `djmoney` is included to `INSTALLED_APPS`.
Since Django 1.7 it is possible to specify startup code in `apps.py`, which is done for `MoneyField`

Stranger6667:
Thank you!

PR #248, opened 2016-12-19
Rules for uploading a release

benjaoming:
@Stranger6667 if you use `make release` for future releases, it will invoke `twine` etc. to upload it.

We could potentially add `git tag -s`, but for my own usecase, I have to tell `git tag` which PGP key to use, so it would fail for me, which is why I'm not using it. It's a manual thing to remember, as with creating the release entry on Github.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,3 +37,11 @@ coverage:
 
 install: clean
 	python setup.py install
+
+release: dist  ## Generate and upload release to PyPi
+	twine upload -s dist/*
+
+dist: clean  ## Generate source dist and wheels
+	python setup.py bdist_wheel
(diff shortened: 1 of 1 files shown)


Stranger6667:
Hm, I don't have a permission for PyPI :)

benjaoming:
Garh, alright, but send me your PyPi username if you want, or if you prefer to keep the responsibilities separated, that's also cool :)

Stranger6667:
My username is Stranger6667 :) Thanks

benjaoming:
Argh, #facepalm - I don't have admin rights, maybe @reinbach has :)

PR #249, opened 2016-12-26
Formfield doesn't use model field's default amount

tsouvarev:
If you have model field with some `default`, then this amount should be set as initial for field in form


Stranger6667:
Excellent! Thank you
I'll add a changelog entry later :)

tsouvarev:
Thanks for a really quick merge, would be nice if you make release for PyPI :)

Stranger6667:
You're welcome :) I think we could release `0.10.1`, what do you think, @benjaoming ?

benjaoming:
@Stranger6667 sounds good and done!

I'm bumping straight to `0.10.2dev0` in master...

PR #251, opened 2016-12-30
Added ability to configure decimal places output

ivanchenkodmitry:
Tests in django master branch are failed because they changed the logic of rendering Select widget.

I decided create pull request in django project to remove '\n' character in select options which is comes from template rendering.
See details here:
https://github.com/django/django/pull/7769

I hope this request will be approved and then django-money tests will be passed.
Otherwise we should fix the test.

Thank you.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -27,7 +27,7 @@ from .._compat import (
     split_expression,
     string_types,
 )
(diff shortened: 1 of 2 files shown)


ivanchenkodmitry:
Hi!
Is it possible to rerun integration tests for the pull request?

Stranger6667:
Hello!
Yes, I restarted it just now

Stranger6667:
Thank you for your contribution!

PR #261, opened 2017-02-02
Added support for nullable ``CurrencyField``. Closes #260

### LINKED ISSUE #260: Ability to specify nullable CurrencyField
It could be useful to have an ability to specify nullable currency column. E.g. pass ``default_currency=None`` to ``MoneyField``. But now it raises an error.

Stranger6667:
I don't know if it is enough. Probably it should be possible to have ``default_currency='EUR'`` and nullable ``CurrencyField``


Stranger6667:
Hi! @benjaoming 
What do you think about the PR? Would it be better to use extra keyword argument to define nullable ``CurrencyField``?
Like:

```
money = MoneyField(..., default_currency='CZK', nullable_currency=True)
```
In current implementation it is not possible to use default currency (which is not None) and create instances with ``currency=None``

benjaoming:
Hi @Stranger6667 !

I'm just looking at this really quickly and haven't been having much issues with django-money, so not sure why nullable currency fields are important..

 - Does it make sense to have amount=123, currency=null? IMO it doesn't make sense but it requires extra validation

 - If amount=null, we can always infer currency=null, regardless of the currency field's actual value

IMO it's nice to be strict about always setting a currency. It makes it easier to enforce proper constraints on the DB.

Stranger6667:
I'm working on analytical application, which requires a lot of straight SQL code and use the DB, which is filled from Django application with django-money inside.

I've faced with situation when I have to calculate total amounts per currency, so I got in the most simple case:
```
# SELECT amount_currency, SUM(amount) from test_currency GROUP BY 1;
 amount_currency | sum
-----------------+-----
 CZK             |  15
 XYZ             |
 USD             |  40
```

It confuses me a bit, because there is nullable amount but some default currency. I think if there is no amount, then it should be really no currency. This default currency could be changed over time, so DB will contain multiple currency values, which should be somehow excluded from aggregation. 
In case with NULL there is always one value to exclude. 

In case if default_currency is something realistic like EUR, then it will slowdown aggregation, unless FILTER clause will be used (PostgreSQL):
```
SELECT amount_currency, SUM(amount) FILTER (WHERE amount IS NOT NULL) FROM test_currency GROUP BY 1;
```
It will not be the case if there will be no null values for EUR currency. SUM is simplest case, the more expensive aggregation, the more penalty will be for passing null values to this aggregate.
In some cases null values could be filter out in WHERE clause, but if there are more aggregates (like count of transactions or anything which requires all rows, even with default currency and amount=NULL) it will not work in this way.

Basically it is my motivation to have an ability to set nullable currency column.

I absolutely agree, that some combinations are non-sense, IMO there should be 2 cases:
1) Money - amount and currency
2) No money - no amount and no currency

But now there is no possibility to set currency as NULL :(

benjaoming:
Problem is with 2)

If you introduce the possibility to have nullable amount AND currency, you have to add validation for amount=123 and currency=null.

We could solve it in the `save` method or `clean` on a python-level.. not a guaranteed DB constraint.. but better than nothing.

Something like?

```python
if self.default is None and self.default_currency is None:
    if self.amount and not self.currency:
        raise ValidationError(_("Cannot set the amount without setting a default currency"))
```

benjaoming:
@Stranger6667 

Have updated this to latest master. Could you consider adding this test case:

```python
    def test_fails_with_null_currency(self):
        with pytest.raises(NotSureWhichException):
            instance = ModelWithNullableCurrency.objects.create(money=10.0)
```

djmoney/models/fields.py
```diff
@@ -174,7 +174,7 @@ def __init__(self, verbose_name=None, name=None,
                  currency_field_name=None, **kwargs):
         nullable = kwargs.get('null', False)
         default = self.setup_default(default, default_currency, nullable)
-        if not default_currency:
+        if not default_currency and default is not None:
```
benjaoming: What concerns me with this is that we allow assigning just number types to a Money field, and we might then have a database value that says `money=1.234` and `currency=null` ?

Stranger6667:
We can ensure integrity in this way:

```SQL
CREATE TABLE test (
    id BIGSERIAL PRIMARY KEY, 
    amount NUMERIC NULL, 
    currency VARCHAR(3) NULL, 
    CONSTRAINT not_null_amount CHECK (
        (amount IS NOT NULL AND currency IS NOT NULL) OR (amount IS NULL AND currency IS NULL)
    )
);
```

But I'm not sure if it is easily possible to add it via Django's ORM directly + not sure if this constraint will work on other RDBMS, but I'll try to find out

Stranger6667:
`CHECK` constraints are ignored by `MySQL` 5.7 :/

https://dev.mysql.com/doc/refman/5.7/en/create-table.html

benjaoming:
@Stranger6667 I think you can add the constraint in Python. But currently, it will accept a float value without a currency, as I've described in the comment. So if this is fixed, I think it's possibly the "least harmful" way to introduce the behavior.. IMO it's better than adding a complex DB-specific constraint.

Stranger6667:
Added a check on Python level. 

Regarding DB-level constraint. 
`CHECK` constraint was in the standard before `SQL-92`, and it is very sad, that it is ignored by MySQL. Even in MySQL 8.0.

So, I'd say that the constraint itself is not really DB-specific, but MySQL is a very specific RDMS :/

Hope, that this will change in the future.

tests/test_models.py
```diff
@@ -328,6 +329,25 @@ def test_shared_currency(self):
...
+        assert instance.money_currency is None
+
+    def test_create_default(self):
+        money = Money(100, 'SEK')
+        instance = ModelWithNullableCurrency.objects.create(money=money)
+        assert instance.money == money
+
+    def test_fails_with_null_currency(self):
+        with pytest.raises(ValueError) as exc:
+            ModelWithNullableCurrency.objects.create(money=10)
```
benjaoming: Awesome! :)

PR #271, opened 2017-02-25
Added ability to set custom currency choices via ``CURRENCY_CHOICES`` settings option. Closes #211

### LINKED ISSUE #211: How to set custom name for currency for dropdownbox?
Hi, 

In the dropdownbox I just want to show USD, EUR, etc. But I couldnt find how to do it.

Instead I changed the djmoney\settings.py into this:

``` python
if PROJECT_CURRENCIES:
    try:
        CURRENCY_CHOICES = [(CURRENCIES[code].code, name) for (code, name) in PROJECT_CURRENCIES]
    except (ValueError, TypeError):
        CURRENCY_CHOICES = [(code, CURRENCIES[code].name) for code in PROJECT_CURRENCIES]
else:
    CURRENCY_CHOICES = [(c.code, c.name) for i, c in CURRENCIES.items() if
                        c.code != DEFAULT_CURRENCY_CODE]
```

And made the CURRENCIES setting a tuple:

``` python
CURRENCIES = (('EUR', 'EUR €'), ('USD', 'USD $'))
```

but is there an official way to set it on a project-level?

Stranger6667:
As was requested in #211 

What do you think about this feature?


benjaoming:
@Stranger6667 I'm unfortunately lacking time and bandwidth for the next 2-3 weeks :/ will put it on my getting-back TODO - if you can wait..

Stranger6667:
@benjaoming no problem :)

benjaoming:
Solid work! :)

tests/test_settings.py
```diff
@@ -1,24 +1,51 @@
...
+class TestCurrencies:
+
+    def assert_test_passed(self, testdir, lines):
+        result = testdir.runpytest_subprocess(
+            '--verbose', '-s',
+            '--ds', 'app_settings',
+            '--cov', 'djmoney',
+            '--cov-config', 'coveragerc.ini',
+        )
+        result.stdout.fnmatch_lines(lines)
```
benjaoming: Does this run all tests again but with different settings?

tests/test_settings.py
```diff
@@ -1,24 +1,51 @@
...
+
+    def test_custom_currencies(self, testdir):
+        testdir.makepyfile(app_settings='''
+        # -*- coding: utf-8 -*-
+        INSTALLED_APPS = ['djmoney']
+        CURRENCIES = ['USD', 'EUR']
+        CURRENCY_CHOICES = [('USD', 'USD $'), ('EUR', 'EUR €')]
+
+        SECRET_KEY = 'foobar'
+        ''')
```
benjaoming: This is tooooooo weird IMO :)

tests/test_settings.py
Stranger6667: It works in the following way:

- We create a new settings file in temp directory
- We run pytest in subprocess to avoid all possible module reloading issues, which will be the same if we will run it against the new project
- test is performed in a regular way
- we compare expected output in the main process

benjaoming: Still a couple of questions :)
 
 * Do you mean *all* tests?
 * Why is module reloading a requirement? Shouldn't such be done explicitly?
 * All tests are subject to "what if modules are loaded in a different order, let's load everything again in an isolated environment"

Can this be used?

https://docs.djangoproject.com/en/dev/topics/testing/tools/#overriding-settings

Stranger6667: 1. No, it runs only tests defined in the temp directory, in these cases it is 1 subprocess test per test method in the class
2. If we load ``djmoney.settings`` in first test once it will load Django's settings. In second test if we need to test another combination of settings we will need to change them in settings.py, remove `djmoney/settings.pyc` and load `djmoney.settings` again to get new values and cover all logic branches in `djmoney.settings`. I'm not sure if we can do it in the same way on all Django/Python combinations that we support. So, this solution is rough but isolated :)
3. Almost, in this case the most problematic thing is that avoiding module reloading and caching issues seems to me core complicated than running every test in the isolated environment.

I'm not sure about [https://docs.djangoproject.com/en/dev/topics/testing/tools/#overriding-settings](https://docs.djangoproject.com/en/dev/topics/testing/tools/#overriding-settings), I'll try it :)

Probably running python with -B option could help. I will try to simplify tests :)

benjaoming: Ah, I see what you mean after reading this in the docs:

> Finally, avoid aliasing your settings as module-level constants as override_settings() won’t work on such values since they are only evaluated the first time the module is imported.

Maybe we can use `override_settings` + `reload(djmoney.settings)` ?

Stranger6667: Yes! It works for Django 1.10 and Python 2.7. I'll check the rest and push changes
Thank you for the review :)
The initial implementation is really overkill

PR #283, opened 2017-03-22
Better money support. Closes #276

### LINKED ISSUE #276: Better Money support
Now we're have some problems, which are related to using of `moneyed.Money`.

### F expressions support is not complete

This is possible:

```python
F('field') + Money(1, 'EUR')
```

But this one will rise ``TypeError``:

```python
Money(1, 'EUR') + F('field')
```

```TypeError: Cannot add or subtract a Money and non-Money instance.```

The ``moneyed`` package should not know about Django's ``F`` expressions, and we need to implement it on the `django-money` side to fix the issue.

### Lacking of clarity about MoneyPatched

IMO the name doesn't clarify the purpose of the class. Sometimes we're using it, sometimes not, but it is not clear why.

### What to do?

Use only `Money` from `djmoney`, not from `moneyed`, which will be a subclass of ``moneyed.Money`` and will contain all additional logic to support Django. 
If we will use this class internally and encourage users to use it, than we will need only one entity instead of two. In this case we can guarantee, that all Django features will be supported by this class, we can't do it for ``moneyed.Money``.
Probably, we should support ``moneyed.Money`` as we do now, but add a note, that in this case some features are not supported.

So, my proposal is:

- Rename ``MoneyPatched`` -> ``Money``
- Add complete ``F`` expressions support
- Replace all usages of ``moneyed.Money`` to ``djmoney.Money``
- Update documentation accordingly

@benjaoming , what do you think?

Stranger6667:


Stranger6667:
With this PR our localization code fully relies on ``moneyed`` implementation, like it suppose to be, I believe. But there are few caveats:

1. To display RUB sign (`руб.`) in en-US locale we have to disable ``USE_L10N``, which seems strange for me, because it should (from my intuition) be exactly opposite -> if you want localized currency sign - set ``USE_L10N`` to True.
2. This behaviour is different in stable and latest versions of ``py-moneyed``, because of [this PR](https://github.com/limist/py-moneyed/pull/75). Some decimal separators were changed, thus there more failed tests in `test_tags.py`.

First issue exists because there are few signs definitions for en-US locale [here](https://github.com/limist/py-moneyed/blob/master/moneyed/localization.py#L379). And if there is no currency sign defined for given locale-currency pair, prefix-suffix fallback will be straightforward - `('', " %s" % currency_code)` (it is [here](https://github.com/limist/py-moneyed/blob/master/moneyed/localization.py#L49))

I want to discuss about these issues. @benjaoming, what do you thing about this? Should we rely on ``py-moneyed`` localization?

benjaoming:
Sorry for the lack of feedback. I agree that 1) is by far the most intuitive understanding of what `USE_L10N` should do :+1: 

Regarding backwards compatibility, and since this would silently introduce a different formatting of money which may have consequences for someone upgrading, it might be better to introduce it with a substantial version bump and release note... ?

Stranger6667:
I added a release notes about this API change, also there is a backward compatibility ``MoneyPatched`` class, thus it should be easier to migrate to the new version.

Stranger6667:
Hi @benjaoming !
I'd like to discuss a couple of problems with you :)

1. Should be force users to use ``djmoney.money.Money`` instead of ``moneyed.Money`` with some deprecation warning?
``djmoney.money.Money`` has full support for ``F`` expressions and Django-dependent localization when ``moneyed.Money`` doesn't and usage of ``moneyed.Money`` might lead to errors :(

2. What do you think about the class path itself ``djmoney.money.Money``? 3 times ``money`` in a single line :( I tried to simplify it to ``djmoney.Money``, but it leads to many local imports inside the ``Money`` class, because otherwise this import leads to load all Django models, etc.

Best Regards

benjaoming:
> 1. Should be force users to use ``djmoney.money.Money`` instead of ``moneyed.Money`` with some deprecation warning?
> ``djmoney.money.Money`` has full support for ``F`` expressions and Django-dependent localization when ``moneyed.Money`` doesn't and usage of ``moneyed.Money`` might lead to errors :(

You could also instatiate ``djmoney.money.Money`` automatically if `type(field) == moneyed.Money` and not warn the user. I don't think warnings are nice in this case, because either then user deliberately wants this type because we have made some bug/lack/regression in our subtype (which we should fix).

If docs are updated, release notes advice to change to the new type, and we automatically instantiate it internally, that would be easiest for everybody, right? We can assume only on type throughout the code..

Would there exist any reason at all to have ``moneyed.Money`` types when a new implementation is adding django-money specific value to the type?

> 2. What do you think about the class path itself ``djmoney.money.Money``? 3 times ``money`` in a single line :( I tried to simplify it to ``djmoney.Money``, but it leads to many local imports inside the ``Money`` class, because otherwise this import leads to load all Django models, etc.

image

:)

Importing `djmoney.money.Money` in `djmoney.__init__` seems to cause a rather complicated stack of modules and packages to be loaded, so I'm not sure that's a good idea.

Stranger6667:
> If docs are updated, release notes advice to change to the new type, and we automatically instantiate it internally, that would be easiest for everybody, right? We can assume only on type throughout the code..

Yes, I'll add changes to change a type for ``moneyed.Money`` instances.

> Would there exist any reason at all to have moneyed.Money types when a new implementation is adding django-money specific value to the type?

I think, no, there would be no reasons to use it - it is almost the same thing, but with Django-related features (except for historically ``__float__`` method), all original logic is the same. Even not supporting "/" operator on Python 2 :)
At this point, original ``moneyed.Money`` instances couldn't be used in some cases, which could lead to errors in migrations or in usage with ``F`` objects.

> Importing djmoney.money.Money in djmoney.__init__ seems to cause a rather complicated stack of modules and packages to be loaded, so I'm not sure that's a good idea.

Agreed :) Let's keep it simple

Stranger6667:
Now the code is working with ``moneyed.Money`` as well :)

Stranger6667:
I tested this PR at my main project, all tests passed. Also, I added some tests here to test the code with ``moneyed.Money``. Everything is fine now :)

benjaoming:
Well done!!

Issue #292, opened 2017-05-15
The POST/PATCH/PUT methods doesn't work with MoneyFields in Django Rest Framework


gonzalobf:
The <fieldname>_currency field is ignored and as result, the default money is always used.

Django version: 1.11
Django Rest Framework version: 3.6.2
django-money: 0.10.2 (and also tried with master branch)

PR #293, opened 2017-05-15
Issue #292: Fix POST/PUT/PATCH in django rest framework. 

gonzalobf:
The currency value is ignored. Therefore, the default currency is always saved and used.
Implementing the `get_value` method in MoneyField class fix the problem.

### DIFF
--- a/djmoney/contrib/django_rest_framework/fields.py
+++ b/djmoney/contrib/django_rest_framework/fields.py
@@ -14,7 +14,7 @@ class MoneyField(DecimalField):
     """
 
     def get_value(self, data):
-        amount = super().get_value(data)
+        amount = super(MoneyField, self).get_value(data)
         currency = data.get('{}_currency'.format(self.field_name), None)
         if currency:
             return Money(amount, currency)
--- a/djmoney/contrib/django_rest_framework/fields.py
+++ b/djmoney/contrib/django_rest_framework/fields.py
@@ -13,6 +13,13 @@ class MoneyField(DecimalField):
     does decimal's validation during transformation to native value.
     """
 
+    def get_value(self, data):
+        amount = super().get_value(data)
+        currency = data.get('{}_currency'.format(self.field_name), None)
+        if currency:
+            return Money(amount, currency)
+        return amount
+
     if IS_DRF_3:
 
         def to_representation(self, obj):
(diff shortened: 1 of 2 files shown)


Stranger6667:
Hello @gonzalobf !
Could you, please add a test case, which demonstrates the issue?
Besides this, please, fix ``super`` calls syntax for Python 2 and add some notes to changelog.

djmoney/contrib/django_rest_framework/fields.py
```diff
@@ -40,6 +47,12 @@ def validate(self, value):
             amount = value.amount if isinstance(value, Money) else value
             return super(MoneyField, self).validate(amount)
 
+        def field_from_native(self, data, files, field_name, into):
+            super(MoneyField, self).field_from_native(data, files, field_name, into)
+            currency = data.get('{}_currency'.format(field_name), None)
```
Stranger6667: Unfortunately, Python 2.6 doesn't support this syntax

gonzalobf: @Stranger6667 Thank you. I will fix it now.

gbarbaten: Hi @Stranger6667 , why are you still supporting Python 2.6 (11 yo, not supported by Django 1.8.18/LTS version) or DRF < 3 (any commit in the last 2 years)?

Stranger6667: Hello @gbarbaten !
There are no specific reasons for supporting old Python versions (as well as old Django & DRF versions) besides it's quite easy thing to do, because, historically, we have a codebase, which works on old Django versions.
I have some thoughts about dropping support for Django < 1.8 & Python 2.6. I think we could drop it after the next release - djmoney 0.11 will be the last version to support Python 2.6 & Django < 1.8.
@benjaoming what do you think about our support policy? :)

benjaoming: I think that sounds good. Django 1.11 will perhaps be important to support for quite some time because it will be the last "gateway" release for people to move from py2 to py3.

Anyways, if someone is dependent on old Django + old Python, then they might as well also use "old djmoney" :)

gonzalobf:
Hi @Stranger6667!

It should be working now. Any feedbacks or suggestions are welcomed :)

Thank you

Stranger6667:
I think, we're done here :) I'll merge the PR, add some notes to the changelog and, hopefully, soon we'll have a new release

PR #303, opened 2017-05-31
0.11.x into master

benjaoming:
Firstly, a clarification that I just released the current `master` branch as `0.11.2` - I see nothing in the log after 9e74fd0dd7297c37738810fe6ff3cc1deee143f9 that hints it had `0.12dev` changes... so it should be without problems. This wasn't really a deliberate action, but I did some changes in `master` when you landed `0.11.1` in `0.11.x` and now the `0.11.2` PR bug fix went into `master` instead of `0.11.x` :)

I know the `0.11.x` branch is needed as soon as `master` diverges from the `0.11` release series, so I didn't delete it (but maybe we should).

Here's what I did:

1. I rebased the current master onto `0.11.x` and force-pushed so it's consistent with `master`.
2. Then I merged `0.11.x` into master and made this PR

I can live with having a branch for supporting the `0.11.x` series, but I have two arguments to back up a branch strategy where `0.11.x` only gets created once needed:

1. To wait with `0.11.x` until something is merged that *cannot* be released in 0.11 and is so unstable that we can't release `0.12` (thus making the `master` branch unreleasable).
2. To wait with `0.11.x` until a patch arrives for an issue that has to be backported and then branch of the latest 0.11 tag (currently `0.11.2`).

Also, I think it's cool to skip backporting fixes to 0.11 series... I mean, that's more because I prefer having proper backwards support in new releases so everyone can always enjoy the latest release of an app :)

### DIFF
--- a/djmoney/__init__.py
+++ b/djmoney/__init__.py
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
 
-__version__ = '0.11'
+__version__ = '0.11.1'
 default_app_config = 'djmoney.apps.MoneyConfig'
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -219,7 +219,7 @@ class MoneyFieldProxy(object):
 
     def __get__(self, obj, type=None):
         if obj is None:
-            raise AttributeError('Can only be accessed via an instance.')
+            return self
         data = obj.__dict__
         if isinstance(data[self.field.name], BaseExpression):
             return data[self.field.name]
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,11 @@
 Changelog
 =========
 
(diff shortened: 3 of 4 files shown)


benjaoming:
*doh* seemed I force-pushed a rebase conflict to `0.11.x` and that's what I'm trying to merge into master now :)

screenshot from 2017-05-31 11-32-50

benjaoming:
Argh, this is kind of madness.. I can't make Github consistently show the difference between force-pushed (fixed) version of 0.11.x. So I created a branch with a merge commit instead.

`0.11.x-orig` and `0.11.x` are identical branches on my local machine

https://github.com/django-money/django-money/compare/master...django-money:0.11.x
https://github.com/django-money/django-money/compare/master...django-money:0.11.x-orig

However, there's a difference according to Github.

PR #316, opened 2017-07-07
Removed unused & deprecated code

Stranger6667:
``choices`` was deprecated few years ago, probably now is the time to finally remove ambiguous argument.

### DIFF
--- a/djmoney/forms/fields.py
+++ b/djmoney/forms/fields.py
@@ -1,8 +1,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from warnings import warn
-
 from django.core.exceptions import ValidationError
 from django.forms import ChoiceField, DecimalField, MultiValueField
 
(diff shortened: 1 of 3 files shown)


Stranger6667:
Also, I removed ``price_field`` argument from ``CurrencyField``, which is unused + I simplified the signature of the field.

benjaoming:
Removing `price_field` can break stuff even if it's unused - not sure if it has ever been used...

Are you sure about removing `choices` and not the other way around?

The warning said:

```
            warn('currency_choices will be deprecated in favor of choices', PendingDeprecationWarning) 
```

But the new constructor params are with `currency_choices`:

```
def __init__(self, currency_widget=None, currency_choices=CURRENCY_CHOICES, max_value=None, min_value=None,
```

Stranger6667:
> Removing ``price_field`` can break stuff even if it's unused - not sure if it has ever been used...

Probably, migration framework will detect changes, but it seems like a private API (we don't use it internally and don't have any usage documentation / mention about this attribute), that we don't expose. It could break some dependent code in theory, but it should be really specific thing. I think, that it is a really, really small possibility to break something with this change, but it still exists.

> Are you sure about removing choices and not the other way around?

I read your comment about it:

```
# choices does not make sense in this context, it would mean we had
# to replace two widgets with one widget dynamically... which is a
# mess. Instead, we let currency_choices be the same as choices and
# raise a warning.
```

May be I get your comment wrong but did you mean that ``choices`` means for choices for ``MoneyField`` as a whole, and ``currency_choices`` means choices only for "currency" part of ``MoneyField``? I'm confused a bit :)

benjaoming:
Sorry about the blahblah :)

If you think about it, specifying choices for a MoneyField would be very strange, because normally it renders 2 widgets like this:

    (NumberInput, CurrencyDropdown)

If we want to limit the choices, it would have to be displayed as just 1 widget. And I think this could lead to some issues in other places.

The original hack was to force displaying a real warning message and then just use the value for something... but maybe it should just be discarded completely and then display a warning that there's no reasonable implementation of `choices` for `MoneyField` for now?

Stranger6667:
> If we want to limit the choices, it would have to be displayed as just 1 widget. And I think this could lead to some issues in other places.

Agree, it makes sense. 

But the ``currency_choices`` is used only in this comparison as a value for the ``choices`` variable and it means, that there are 2 variables that serve the same purpose. I believe, that it would be better to use only one variable for choices of the currency drop-down widget.

Thus being said, we could differentiate two things:

- choices - values, that are available for the multi-field. E.g. - 10 EUR, 20 EUR, 30 EUR, etc. In this case, we have to treat MoneyField as a single widget, which is a kinda different thing. (1)
- currency_choices - values for currency field only. (2)

What do you think? 

> but maybe it should just be discarded completely and then display a warning that there's no reasonable implementation of choices for MoneyField for now?

I don't know if we have to display a warning for using ``choices`` because we don't have the feature (1) at all as well as there is no ``choices`` argument for ``MultiValueField``, so it is a django-money extension, which always worked as choices for ``CurrencyField``.

benjaoming:
Sorry, I completely forgot about getting this in for 0.13

0.14 ?

Stranger6667:
Sure, I'll update ``changelog.rst`` in this branch

PR #341, opened 2017-10-16
[WIP] Fix querying on related models. Fixes #327

### LINKED ISSUE #327: Filtration on related models doesn't respect currency
Lets use ``ModelRelatedToModelWithMoney`` from our test suite to build this query:

```ModelRelatedToModelWithMoney.objects.filter(moneyModel__money__lt=Money(15, 'USD')).only('id')```

Actual SQL (I use "model" instead of "testapp_modelrelatedtomodelwithmoney" and "vanilla" instead of "testapp_modelwithvanillamoneyfield" for readability):
```SQL
SELECT "model"."id" FROM "model" INNER JOIN "vanilla" ON ("model"."moneyModel_id" = "vanilla"."id") WHERE "vanilla"."money" < 15
```

Expected SQL:
```SQL
SELECT "model"."id" FROM "model" INNER JOIN "vanilla" ON ("model"."moneyModel_id" = "vanilla"."id") WHERE "vanilla"."money" < 15 AND "vanilla"."money" = 'USD'
```

Stranger6667:
Thoughts:

Patching should be done after all models are loaded, and then the models' graph should be traversed and appropriate managers should be patched. 

Basically, all models that could be reached from some model with money should be patched because the money model could be queried from them. So, the thing is to find all strongly connected components containing at least one money model and patch all models in these components.

Related algorithm: Kosaraju SCC

I'm not sure if bringing this kind of complexity worth it, but it should solve the problem. Anyway, I already implemented the algorithm a few months ago, so it could be adapted for this purpose.

Alternatively, we can patch all managers

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -29,6 +29,7 @@ Fixed
 - Fixed ``F`` expressions support for ``in`` lookups. `#321`_ (`Stranger6667`_).
 - Fixed money comprehension on querysets. `#331`_ (`Stranger6667`_, `jaavii1988`_).
 - Fixed errors in Django Admin integration. `#334`_ (`Stranger6667`_, `adi-`_).
+- Fixed respecting currency in filtration on related models. `#327`_ (`Stranger6667`_).
 
 Removed
 ~~~~~~~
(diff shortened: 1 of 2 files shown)


benjaoming:
References #327

Stranger6667:
It seems like I won't finish this PR - but the problem may be fixed in the future still.

PR #344, opened 2017-10-27
Fix example for `CURRENCY_CHOICES`

myii:
Existing example results in:

```python
AttributeError: 'tuple' object has no attribute 'sort'
```

This error and solution has already been mentioned by @ChessSpider at https://github.com/django-money/django-money/issues/211#issuecomment-303083655.


Stranger6667:
Thank you!

PR #349, opened 2017-11-13
Replace field.rel.to with field.remote_field.model

richardowen:
`Field.rel` and `Field.remote_field.to` are removed in Django 2.x.
This fixes compatibility with Django 2.x, prevents warnings being emitted on 1.10 & 1.11 and keeps backwards compatibility with 1.8.


Stranger6667:
Hi! Thank you for the contribution!

PR #357, opened 2017-11-20
0.12.x into master + version bump master

benjaoming:


benjaoming:
Django master builds are breaking, but that seems more likely to be because of errors in Django?

Stranger6667:
Yeah, ``QUERY_TERMS`` constant is finally removed.

benjaoming:
@Stranger6667 - will let the builds break, a separate fix is required to make the `master` work again..

Issue #371, opened 2017-12-12
BaseMoneyValidator fails to validate falsy ints and Decimals


1337:
In [the validator code](https://github.com/django-money/django-money/blob/master/djmoney/models/validators.py#L32), any subclass of `BaseMoneyValidator` will short-circuit the validation if the validator was initialised with a falsy integer or or decimal.

Attach a debugger at this code:

    if not limit_value:  # 0 or Decimal('0')
        return

Steps to reproduce:

1. Create a `MoneyField(max_digits=10, decimal_places=2, default_currency='USD')` (copied from example) but with an additional validator, like so:

    ```
    MoneyField(
        max_digits=10, decimal_places=2,
        default_currency='USD',
        validators=[
            MinMoneyValidator(0),
        ]
    )
    ```

1. Make a migration and migrate.
1. Run the server and save any object with that new field with the Money equivalent of `(-1, 'USD')`. Observe that the validator does not raise a `ValidationError`.

This affects the common use case of validating against negative amounts of money.

Stranger6667:
Hello @1337 !
Thank you very much for the report and reproducing steps :) I'll come back to this issue tomorrow and push the fix

Kind regards

Stranger6667:
Wow, when I was typing the reply you pushed the fix :) Cool

1337:
It's my first time using tox so please check it out! Meanwhile I'll make a PR for it.

PR #372, opened 2017-12-12
Fix BaseMoneyValidator with falsy limit values

1337:
Fixes issue #371

* `MinMoneyValidator(0)` must allow 0, 1, and disallow -1 of a currency.
* `MaxMoneyValidator(0)` should also allow 0, and only other negative amounts of currency.

Made some forms and models that seem redundant. Please let me know if there are better ways to go about it.

### DIFF
--- a/djmoney/models/validators.py
+++ b/djmoney/models/validators.py
@@ -29,7 +29,7 @@ class BaseMoneyValidator(BaseValidator):
     def __call__(self, value):
         cleaned = self.clean(value)
         limit_value = self.get_limit_value(cleaned)
-        if not limit_value:
+        if limit_value is None:
             return
         if isinstance(limit_value, (int, Decimal)):
             cleaned = cleaned.amount
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -6,7 +6,7 @@ Changelog
 `Unreleased`_
 -------------
 
-- WIP
+- Fixed `BaseMoneyValidator` with falsy limit values. `#371`_ (`1337`_)
 
 
 `0.12.2`_ - 2017-12-12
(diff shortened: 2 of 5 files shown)


Stranger6667:
LGTM!
Very nice PR! Thank you for your work :)

I have a few notes on it.
Please, apply isort (via ``tox -e fix-lint``) and update the changelog in the following way:

- Add yourself at the bottom of changelog.rst:

```
.. _1337: https://github.com/1337
```

- Add a link to the issue a bit upper:

```
.. _#371: https://github.com/django-money/django-money/issues/371
```
- Add a note about the change like this to the "Unreleased" section:

```
Fixed
~~~~~

- <Your change not about the fix>. `#371`_ (`1337`_)
```

1337:
Like this? (see above)

Stranger6667:
Exactly! :) Thanks

Stranger6667:
Hi @benjaoming!
We talked today about routine merging, but here it will not work as I see. Because initial base branch was master :( Seems like cherry picking is the way to go here for porting it to ``0.12.x``, since it is not possible to change the base branch of the PR author.

What do you think?

Kind regards

benjaoming:
@Stranger6667 yes, you are right - one way would be to change the branch and fix merge conflicts with Githubs interface.

In case it seems like an overload, then cherry-picking would be the answer!

A third option would be to stop having a `0.13dev` brewing in `master`.. maybe it's not relevant?

benjaoming:
Ah okay, disregard, wasn't aware that `master` is already a considerable number of commits ahead of 0.12.x

benjaoming:
cherry-pick seems like the way

PR #375, opened 2017-12-18
Drop Django 1.10 support

Stranger6667:


tests/migrations/helpers.py
```diff
@@ -19,7 +19,7 @@ def makemigrations():
     questioner.input = lambda x: 'y'
 
     os.system('find . -name \*.pyc -delete')
-    if VERSION >= (1, 10):
+    if VERSION >= (1, 11):
```
benjaoming: This is not the same as "dropping" support, isn't this more like breaking it? :)

benjaoming:
Since I'm someone who's running Django 1.10 with django-money, I have an opinion here :)

I don't think this PR is simply dropping the support, it looks like it will break it, too?

I think the idea of supporting something is just whether or not, we should both with test failures?

So in case a test fails or something else break with Django 1.10: Too bad, it's not supported anyways. But until then, installing via `pip install django-money` should not force Django to upgrade, too.

benjaoming:
Now I have upgraded the site in question to Django 1.11 :)

tests/migrations/helpers.py
Stranger6667: You're right, I didn't think about it :(

Stranger6667:
Hello @benjaoming !

> I think the idea of supporting something is just whether or not, we should both with test failures?

Could you, please, explain. What did you mean? 

I thought, that we should remove test cases for unsupported Django version. But also I removed all mentions of unsupported Django versions. 
What do you think about it? It kinda forcing users to update Django. May be we should just remove related builds, all shims for unsupported version only in tests and all bug reports / issues for this version?

Kind Regards

benjaoming:
Sorry for the sloppy writing.

What I mean is that the support shouldn't be removed deliberately IMO -- but once something breaks, instead of fixing it, it would make sense for sure to remove it from the test matrix and announce that support has been dropped.

But if the tests are working, doesn't it make sense to just keep saying that Django 1.10 is supported? And when we know it's broken, it makes sense to remove it?

I think it's just as fine to get rid of it, if you see a reason for that or want to have the code maintained :)

It was just a POV because I was using Django 1.10.. now I don't use it anymore :)

benjaoming:
Seems I slowed this down -- so instead it's now merged :)

Stranger6667:
Hello @benjaoming !
Sorry for not getting back to you. 

Since this PR breaks backward compatibility with Django 1.10, the next version should probably be 0.13. Or I could change some lines to make it a non-breaking change. Probably it will allow some users to get the latest version of ``django-money`` without updating Django. But on the other side, 1.10 is EOL already. What do you think? :)

benjaoming:
Yes, I think we are already working on 0.13 in the current `master`? At least 0.12 has broken off into a release branch, so should be safe to make that assumption now.

I liked your release notes regarding the change. And I'm getting into a mode of thought / rationale that's kind of like this: It's too difficult to keep support for older Django releases and develop new features at the same time. It's necessary to make a choice.

PR #382, opened 2018-03-12
[Update money.py]

Robertino10:
"[Robertino10]"
 	 
 	 
"[ def convert_money(value, currency):	 def convert_money(value, currency):]"


Robertino10:
"[Robertino10]"

"[ def convert_money(value, currency):	def convert_money(value, currency):]"

PR #384, opened 2018-03-25
Fix #347

### LINKED ISSUE #347: __float__ method makes numpy/pandas forget currency
```
>>> pandas.DataFrame([moneyed.Money()]).sum()
0    0.00 XYZ
dtype: object

>>> pandas.DataFrame([djmoney.money.Money()]).sum()
0    0.0  # <-- currency missing
dtype: float64
```

This is due to implementing the `__float__` method:

```py
class MM(moneyed.Money):
    def __float__(self):
        return float(self.amount)
```

```
>>> pandas.DataFrame([MM()]).sum()
0    0.0  # <-- currency missing
dtype: float64
```

I'm not saying the `__float__` method should be removed/changed because I'm not sure of what use it is. Just wanted to leave a note about this here.

jonashaag:
@benjaoming I guess this is the only way to fix #347 .

### DIFF
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -1,4 +1,6 @@
 # coding: utf-8
+import warnings
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import F
@@ -23,6 +25,8 @@ class Money(DefaultMoney):
     use_l10n = None
 
     def __float__(self):
+        warnings.warn("float() on a Money object is deprecated. Use the "
+                      "'amount' attribute instead.", DeprecationWarning)
         return float(self.amount)
 
     def __add__(self, other):


benjaoming:
I would tend to agree with this.

Especially because the current implementation doesn't contain any reasoning. But I know that I originally added this... but like 5 years ago :) https://github.com/django-money/django-money/pull/16

But it will be fairly disruptive to people assuming that you can somehow treat Money as floats and then re-add the currency afterwards.

So I would suggest that this only gets merged with:

1) A release note
2) A [minor](https://semver.org/) version bump

jonashaag:
Shouldn't this be a major release as it's backward incompatible? It wasn't documented but it was tested (which one might argue is a kind of documentation)

If we're strict in terms of semver, we'll need a new minor release with the feature deprecated, and a new major with the feature removed somewhen in the future.

benjaoming:
> If we're strict in terms of semver, we'll need a new minor release with the feature deprecated, and a new major with the feature removed somewhen in the future.

Yse, definitely agreed!

benjaoming:
The warning should remind us to finally get this removed, but I'll reopen the issue just to make sure :)

PR #391, opened 2018-04-07
Rounding support via ``round`` built-in function on Python 3.

Stranger6667:
Hi @benjaoming !
Recently I found [this](https://stackoverflow.com/questions/49315079/how-to-round-up-django-money) question on StackOverflow. It could be a feature for ``django-money``.

With this patch, it will be possible to round ``Money`` instances with built-in ``round`` function:

```python
>>> value = Money('1.69', 'USD')
>>> round(value, 1)
<Money: 1.7 USD>
```

Unfortunately, there is no control for rounding on Python 2 - it will use deprecated ``__float__`` method :(

Could we use it like Python3-only feature?

### DIFF
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -1,6 +1,5 @@
 # coding: utf-8
 import warnings
-from decimal import Decimal
 
 from django.conf import settings
 from django.db.models import F
@@ -76,8 +75,8 @@ class Money(DefaultMoney):
     def __html__(self):
         return mark_safe(avoid_wrapping(conditional_escape(self.__unicode__())))
 
-    def __round__(self, n=0):
-        amount = self.amount.quantize(Decimal(10) ** -n)
+    def __round__(self, n=None):
+        amount = round(self.amount, n)
         return self.__class__(amount, self.currency)
 
 
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -1,5 +1,6 @@
 # coding: utf-8
 import warnings
+from decimal import Decimal
 
 from django.conf import settings
 from django.db.models import F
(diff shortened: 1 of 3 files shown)


benjaoming:
Maybe you are scratching the surface of becoming a python3-only project?

Could bump to 1.x and say it's py3-only from there on?

Stranger6667:
No, no, don't get me wrong :) I have no intention of making django-money a Python3-only project (it is kind of too early for now IMO).
Some features just can't work on Python 2, but still could be useful for Python 3 users. Honestly, I'm ok with such features, since we already have Python3-only feature - integer division doesn't work for `Money` instances on Python 2.
This rounding feature changes nothing for Python 2.

Also, you raised an interesting topic about bumping to 1.0 :) I'm really happy, that the project is stable for a long time, and probably, we just need to decide when to that. I was thinking about a roadmap like this:

0.13.2 - the next bugfix release, if needed
1.0 - removing all deprecations like `__float__` method, #316, `MoneyPatched`. Dropping Django 1.8 support?
1.1 - the next release with new features (for example caching for exchange rates)
...
2.0 - Dropping support for Python 2.

What do you think?

Regards,

benjaoming:
@Stranger6667 

Since Python 2.7 is supported by "old" releases like the current 0.13 branch, I think it's entirely okay to drop support moving forwards. If someone runs Python 2.7, it's probably not with the expectation that it will gain new features from following the upgrade paths of dependencies.. I'd rather expect them to not even follow upgrades.

Django 2.0 dropped Python 2, so I think it's okay to follow...

benjaoming:
Note that different rules may apply to rounding money... don't read this :)

https://en.wikipedia.org/wiki/Cash_rounding

PR #396, opened 2018-04-10
Celery integration for exchange rates update. Closes #394

### LINKED ISSUE #394: Documentation about periodic rates updates
Add a new section in "exchange rates" block about how to set up periodic rates updates. Code snippets could look like this:

```python
# tasks.py
from celery import shared_task

from djmoney import settings

@shared_task
def update_rates(backend=settings.EXCHANGE_BACKEND, **kwargs):
    backend = import_string(backend)()
    backend.update_rates(**kwargs)


# settings.py
CELERYBEAT_SCHEDULE = {
    'update_rates': {
        'task': 'path.to.task',
        'schedule': crontab(minute=0, hour=0),
        'kwargs': {}  # For custom arguments
    }
}
```

Stranger6667:


benjaoming:
@Stranger6667 I would say a few things about this :) My opinion is because in django-wiki, we also had these weak sort of integrations, which then broke and there was no one around to fix it, so it just wasted people's time who tried getting it to work. I know this one comes with tests, but if they break, then someone would be challenged having to fix it in some other context.

1. The integration itself is very simple and most people could write it by reading celery docs
1. The footprint on tests is substantive because tests now rely on celery and all its dependencies
1. Celery is very very complex, I would not recommend someone with a need to run a job once every hour or day to use celery for that.

What I would recommend would be to move the integration over to the docs in a special section about `rates` and integration. There, we could also describe normal cronjobs, which IMO are a much cleaner way to implement this.. I would never install and rely on celery just for this :)

If someone wants to use celery for this kind of task, they most likely can benefit from just having a snippet of code to copy and adapt?

Stranger6667:
This makes sense, I'll create a new sub-section in docs and describe options with cron & Celery. So, we can issue a new release :)

benjaoming:
Thanks for understanding @Stranger6667 :smile:  there's nothing wrong with the celery integration itself, but naturally, celery and dependencies will change down the road :)

PR #397, opened 2018-04-10
Remove mentions of unsupported Django 1.10

Stranger6667:
Since Django 1.10 is unsupported since December 2, 2017 I think we can remove it mentions from the README.

Regarding an issue from the end of the page - it was fixed in Django 1.10.2 (released on October 1, 2016), I think we could also remove it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ http://code.google.com/p/python-money/
 
 This version adds tests, and comes with several critical bugfixes.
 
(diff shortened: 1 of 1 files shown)


benjaoming:
Agreed! We just leave the door open for installations that already use 1.10 (as per setup.py requirements `>=1.8`), but don't mention it explicitly. It's already removed from tox.ini, so we don't really know if it works anymore :)

PR #401, opened 2018-04-12
Managers patching for models with custom ``Meta.default_manager_name`…

### LINKED ISSUE #400: Default manager is not patched in some cases
If default manager is customized via ``Meta.default_manager_name``, and the model has ``MoneyField``m then the manager will not be patched, because we only look for ``objects`` manager name.

Stranger6667:
…`. Fixes #400

### DIFF
--- a/djmoney/_compat.py
+++ b/djmoney/_compat.py
@@ -57,7 +57,7 @@ def setup_managers(sender):
     else:
         sender.copy_managers([
             (_id, name, money_manager(manager))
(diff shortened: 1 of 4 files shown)


benjaoming:
@Stranger6667 this is failing tests? But seems important to have included in 0.13.2 :)

Stranger6667:
The test should be skipped on Django 1.8, because the setting was introduced in Django 1.10. I'll fix it

Stranger6667:
Should work now :)

PR #406, opened 2018-05-12
Added djmoney.money.Money.__round__

ZuluPro:
There's a failing test but not due to my patch.

### DIFF
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -29,6 +29,9 @@ class Money(DefaultMoney):
                       "'amount' attribute instead.", DeprecationWarning)
         return float(self.amount)
 
+    def __round__(self, ndigits):
+        return self.amount.__round__(ndigits)
+
     def __add__(self, other):
         if isinstance(other, F):
             return other.__radd__(self)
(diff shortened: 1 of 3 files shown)


Stranger6667:
Hi @ZuluPro !
On the first glance, it seems like a duplicate of #391. I'll take a look today

Stranger6667:
And thank you for the interest :)

Stranger6667:
Aha, it returns a numeric value, instead of `Money` instance. I was thinking about rounding `Money` instances, which seems more logical to me - all math operations should return instances of the same type.

djmoney/money.py
```diff
@@ -29,6 +29,9 @@ def __float__(self):
                       "'amount' attribute instead.", DeprecationWarning)
         return float(self.amount)
 
+    def __round__(self, ndigits):
+        return self.amount.__round__(ndigits)
```
benjaoming: This returns a different type, so will prefer the implementation from #391

benjaoming:
Thanks @ZuluPro - hoping that it's cool we just close this one and fast-track #391 for release... will get it out right meow! :smile_cat:

ZuluPro:
NP men !

PR #410, opened 2018-05-12
Real isolated tests, drop Python 3.3 and update pre-commit hooks

benjaoming:
By switching to `usedevelop=false`, we can avoid some common pitfalls of having an untested distributed, resulting in successful tests on development environments ending up broken in releases.

~I do not, however, understand these new test failures, @Stranger6667 do you have any hints?~ (fixed, was a typo)

Resources:
http://pytest-django.readthedocs.io/en/latest/managing_python_path.html
http://tox.readthedocs.io/en/latest/config.html
https://docs.pytest.org/en/latest/goodpractices.html


benjaoming:
@Stranger6667 I don't think that Python 3.3 was ever a thing? :)

Stranger6667:
Django 1.8 supports 3.3 as far as I remember and we’re supporting Django 1.8 :) probably we could drop Django 1.8 support as well, what do you think?

benjaoming:
It's more that Python 3.3 was never widely used... really, _at all_! No one uses it :)

Django 1.8 is still LTS, but yeah could drop it I guess :)

Stranger6667:
Yes, you're right :) I'll look for some shims for Python 3.3, that could be removed later

PR #413, opened 2018-05-19
Fix packaging

Stranger6667:
Hi @benjaoming ! 

I didn't realize, that we dropped support for Python 3.3 in the `0.13.x` branch. To keep `0.13.x` branch without backward incompatible changes, could we restore it here and cherry-pick the commit to the master branch? Then we can release 0.13.4 bugfix release and continue with cleanup in master. 
Or something like:
- Fix packaging in `0.13.x`
- Merge `0.13.x` to master
- Restore support for Python 3.3 in `0.13.x`
- Release `0.13.4`

What do you think?

Kind Regards

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -6,12 +6,15 @@ Changelog
 `0.14`_ - unreleased
 --------------------
 
+Fixed
+~~~~~
+
+- Packaging of ``djmoney.contrib.exchange.management.commands``. `#412`_ (`77cc33`_, `Stranger6667`_)
+
 Removed
 ~~~~~~~
 
 - Support for Python 3.3 `#410`_ (`benjaoming`_)
-
-
 
 `0.13.3`_ - 2018-05-12
 ----------------------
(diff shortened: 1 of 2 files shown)


benjaoming:
> I didn't realize, that we dropped support for Python 3.3 in the 0.13.x branch. To keep 0.13.x branch without backward incompatible changes, could we restore it here and cherry-pick the commit to the master branch?

Oh oops, the PR where this happened was originally targeted for 0.13.x but didn't drop Python 3.3 support, so I didn't notice later on. Thanks! Will overwrite the branch, since this commit was never tagged or released.

benjaoming:
@Stranger6667 `0.13.x` branch is now restored to 0.13.3 release.

benjaoming:
...but oh lordy, I think now you have to sort out the fix such that it isn't based off the wrong old 0.13.x branch :)

Stranger6667:
Sure, I'll try pick it :)

PR #414, opened 2018-05-19
Fix packaging

benjaoming:
#413

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,16 @@
 Changelog
 =========
 
+
+`0.13.4`_ - 2018-05-19
+----------------------
+
+Fixed
+~~~~~
+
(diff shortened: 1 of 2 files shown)


docs/changes.rst
```diff
@@ -3,6 +3,16 @@
 Changelog
 =========
 
+
+`0.13.4`_ - 2018-05-12
```
Stranger6667: It should be `2018-05-19`

benjaoming: Fixed

Stranger6667:
Probably, we could just ignore the failure or mark it with `allow_failures` on Travis

benjaoming:
Yeah, let's just ignore it and move on. This should be the final 0.13.x release, and since this branch would retain (useless) Python 3.3 support, it doesn't matter that Tox fails to install latest wheel requirement.

PR #415, opened 2018-05-19
Tests changes

Stranger6667:
I'll arrange porting the rest of commits (for dropping Python 3.3 support) in a separate PR


benjaoming:
I already separated those into a `0.14.x` branch, so it's all cool, thanks for this though. I'll keep the `master` aligned to 0.13.x for the 0.13.4 release, but after that perhaps we should make master track the 0.14 series.

Stranger6667:
Thanks :) np

PR #417, opened 2018-05-19
Add djmoney.__init__ to dist

benjaoming:


setup.py
```diff
@@ -40,7 +40,7 @@ def run_tests(self):
     url='https://github.com/django-money/django-money',
     maintainer='Greg Reinbach',
     maintainer_email='greg@reinbach.com',
-    packages=find_packages(include=['djmoney.*']),
+    packages=find_packages(include=['djmoney', 'djmoney.*']),
```
Stranger6667: :((
I missed it :(

benjaoming: That mistake won't happen again in 0.14 series :)

Issue #418, opened 2018-05-24
Same currency conversion getting MissingRate exception


humrochagf:
Hi, I was running a code that sums some values at different currencies and then i found out that calling `convert_money` to the same currency of the source gets a MissingRate exception:

```
>>> convert_money(Money(1, 'USD'), 'USD')
*** MissingRate: Rate USD -> USD does not exist
```

humrochagf:
I found the problem,

When we call `convert_money` in that way inside the `get_rate` function the source and target have different types, `Currency` and `text_type`.

So to fix that i cast them both to `text_type` from `_compat` in order to work across both python 2 and 3 to compare their currency codes.

PR #419, opened 2018-05-24
Fix issue #418 of Same currency conversion

humrochagf:
- Add django-reversion missing dependency for start testing at setup.py test_requirements
- Create tests for this issue
- Cast the currencies to its representation form that gets the currency code and ensure that they are the same

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ test_requirements = [
     'pytest-django',
     'pytest-pythonpath',
     'pytest-cov',
+    'django-reversion',
 ]
 
 extras_requirements = {
--- a/djmoney/contrib/exchange/models.py
+++ b/djmoney/contrib/exchange/models.py
@@ -3,6 +3,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.db import models
 from django.utils.module_loading import import_string
 
+from djmoney._compat import text_type
 from djmoney.settings import EXCHANGE_BACKEND
 
 from .exceptions import MissingRate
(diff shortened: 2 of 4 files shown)


Stranger6667:
Hi @humrochagf !
Thank you for the bug report and PR! It is really appreciated :)
Could you, please, add a changelog entry?

humrochagf:
Done :)

setup.py
```diff
@@ -32,6 +32,7 @@ def run_tests(self):
     'pytest-django',
     'pytest-pythonpath',
     'pytest-cov',
+    'django-reversion',
```
ZuluPro: Why Django-Reverse ? I never use it and there's link to that in Django-Money.

humrochagf: Its a test requirement, does not install when you normally pip install. Just if you install in test mode

humrochagf: If you follow the link above you will see that some tests rely on django-reversion to work: https://github.com/django-money/django-money/search?q=reversion&unscoped_q=reversion

When i first ran the tests, before adding any code they broke because of that non declared dependency. So i added it to the `test_requirements` so the next developers that might contribute to this project does not neet to figure it out :)

But that`s nothing to worry about, neither of these packages are installed with a regular pip install:

```
test_requirements = [
    'mock',
    'pytest>=3.1.0',
    'pytest-django',
    'pytest-pythonpath',
    'pytest-cov',
    'django-reversion',
]
```
To install them you need to run `pip install -e .[tests]` after cloning the repo to your local machine.

benjaoming:
Thanks for this!

Do you need it for 0.13.x release branch or is it fine if we release it in 0.14 instead? @humrochagf

humrochagf:
It can wait, my production code was redirected to a local snapshot branch until it does not become live

PR #421, opened 2018-06-03
Caching of exchange rates. Closes #398

### LINKED ISSUE #398: Cache for currency exchange rates
Default Django cache would not work, since it doesn't allow any args / kwargs by default. Possible options:
1. Use `functools.lru_cache`. Simple to add, but will work only for current process
2. Implement args / kwargs processing for `get_rate` and use default Django cache as a storage. The idea is the same as for `functools.lru_cache` except for LRU part and this storage will be shared among all processes running the same app.

Stranger6667:
Initially, I implemented a decorator-based approach, but the implementation was quite complex for one single function, so I changed it to more straightforward one.


tests/contrib/exchange/test_model.py
```diff
@@ -34,6 +35,15 @@ def test_string_representation(backend):
     assert str(backend) == backend.name
 
 
+@pytest.mark.usefixtures('simple_rates')
+def test_cache():
+    with patch('djmoney.contrib.exchange.models._get_rate', wraps=_get_rate) as original:
+        assert get_rate('USD', 'USD') == 1
+        assert original.call_count == 1
+        assert get_rate('USD', 'USD') == 1
+        assert original.call_count == 1
```
benjaoming: :+1:

djmoney/settings.py
```diff
@@ -32,3 +32,4 @@
 FIXER_ACCESS_KEY = getattr(settings, 'FIXER_ACCESS_KEY', None)
 BASE_CURRENCY = getattr(settings, 'BASE_CURRENCY', 'USD')
 EXCHANGE_BACKEND = getattr(settings, 'EXCHANGE_BACKEND', 'djmoney.contrib.exchange.backends.OpenExchangeRatesBackend')
+RATES_CACHE_TIMEOUT = getattr(settings, 'RATES_CACHE_TIMEOUT', 600)
```
benjaoming: It's cool, but leaves a bit of a desire to start prefixing the settings var names :)

benjaoming:
Hope the hackathon weekend went well :)

PR #428, opened 2018-06-27
Support for indirect rates conversion

Stranger6667:


djmoney/contrib/exchange/models.py
```diff
@@ -64,17 +55,20 @@ def get_rate(source, target, backend=None):
...
+    rates = Rate.objects.filter(currency__in=(source, target), backend=backend)
+    if not rates:
         raise MissingRate('Rate %s -> %s does not exist' % (source, target))
+    # Direct rate
+    if len(rates) == 1:
+        rate = rates[0]
+        if text_type(rate.currency) == text_type(target):
+            return rate.value
+        return 1 / rate.value
+    # Indirect rate
```
benjaoming: So an "indirect rate" means to convert from rate A to rate C by first converting to rate A to B, then B to C? Maybe a few comments in this code to clarify what's going on :)

djmoney/contrib/exchange/models.py
```diff
@@ -64,17 +55,20 @@ def get_rate(source, target, backend=None):
...
+    first, second = rates
+    if text_type(first.currency) == text_type(target):
+        first, second = second, first
```
benjaoming: Wondering: Why this swap?

docs/changes.rst
```diff
@@ -6,6 +6,11 @@ Changelog
 Unreleased
 ----------
 
+Added
+~~~~~
+
+- Support for indirect rates conversion. `#425`_ (`Stranger6667`_, `77cc33`_)
```
benjaoming: Can this be said somewhat more precisely?

> Support for indirect rates conversion through maximum 1 extra step (when there is no direct conversion rate: converting by means of a third currency for which both source and target currency have conversion rates)

djmoney/contrib/exchange/models.py
Stranger6667: There is no guarantee, that the queryset will be ordered somehow, so the first element could be for target currency or for source currency. So, instead of adding an expression to `order_by`, that will put the source currency in the first place I decided to add an extra check here and swap items if they are ordered not as expected

djmoney/contrib/exchange/models.py
Stranger6667: For example, we want to convert NOK to SEK and we have 2 rates - NOK -> USD and SEK -> USD. 
E.g. 1 USD = 7.84 NOK and 1 USD = 8.37 SEK -> 7.84 NOK = 8.37 SEK -> 1 NOK = 8.37 / 7.84 SEK

For the case when we need to convert NOK to USD, we just don't need the second item. Seems like "direct" and "indirect" terms are confusing, I'll try to rephrase it :)

Stranger6667:
Fixed a bug in this code + refactored it :)

djmoney/contrib/exchange/models.py
```diff
@@ -80,6 +80,9 @@ def _try_to_get_rate_directly(source, target, rate):
 
 def _get_rate_via_base(rates, target):
     """
+    :param: rates: A set/tuple of two base Rate instances
+    :param: target: A Rate instance that we want a conversion rate for via rates
```
Stranger6667: it is a string instance

PR #432, opened 2018-07-12
Update README.rst

andytwoods:
Might save the next self-taughter a few hours banging their heads against the screen.


benjaoming:
Thanks. I hope your head is okay ;)

It isn't mandatory to use 4 digits for all, and it may lead to other problems. So I would only recommend this, if you know exactly why you want the extra precision and what to use it for, and how it's being used (rounded etc). There's no point adding it without knowing why.

benjaoming:
Does this sound like okay guidance to you @Stranger6667 ?

PR #442, opened 2018-07-23
Fix validation of ``djmoney.forms.fields.MoneyField`` 

### LINKED ISSUE #439: Form valiadion error on MoneyField
When I tried to create `ModelForm` based on my model containing `MoneyField`, I got error `Enter a list of values.` for the appropriate field.

After some debugging, it seems that problem is in that `djmoney.money.Money` is not instance of `list` or `tuple`, which creates a problem at [clean method of MultiValueField](https://github.com/django/django/blob/master/django/forms/fields.py#L1009).

Simple solution is to override clean method of `MoneyField`, but it would be great if you could address this issue in `django-money` upstream.

Stranger6667:
when ``disabled=True`` is passed to it. Fixes #439

### DIFF
--- a/djmoney/forms/fields.py
+++ b/djmoney/forms/fields.py
@@ -58,7 +58,8 @@ class MoneyField(MultiValueField):
         return super(MoneyField, self).clean(value)
 
     def has_changed(self, initial, data):  # noqa
-        if self.disabled:
+        # Django 1.8 has no 'disabled' attribute
+        if hasattr(self, 'disabled') and self.disabled:
             return False
         if initial is None:
             initial = ['' for _ in range(0, len(data))]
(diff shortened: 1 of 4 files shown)


djmoney/forms/fields.py
```diff
@@ -51,7 +52,15 @@ def compress(self, data_list):
...
 
+    def clean(self, value):
+        if isinstance(value, MONEY_CLASSES):
+            value = (value.amount, value.currency)
+        return super(MoneyField, self).clean(value)
+
     def has_changed(self, initial, data):  # noqa
+        # Django 1.8 has no 'disabled' attribute
+        if hasattr(self, 'disabled') and self.disabled:
+            return False
```
benjaoming: Django 1.8 LTS was actually deprecated as of April 1, 2018: https://www.djangoproject.com/download/

So I should perhaps release this as 0.14.2? I presume nothing experimental entered `master` since 0.14.1 :)

Stranger6667: Yes, thank you :) We'll drop Django 1.8 in 0.15 :)

benjaoming: It's released :)

PR #450, opened 2018-12-17
[WIP] Fixes implicit default value for MoneyField

### LINKED ISSUE #411: Implicit default value set on non-nullable MoneyFields where none is expected
It seems that an unexpected implicit default value is being set on all non-nullable MoneyField values.

When adding the following field to an existing model:

```python
    tax_per_adult = MoneyField(_("price per adult"), decimal_places=5, max_digits=10)
```

…the following is output in Django's auto-generated migration:

```python
        migrations.AddField(
            model_name='flightsubbooking',
            name='tax_per_adult',
            field=djmoney.models.fields.MoneyField(decimal_places=5, default=Decimal('0.0'), max_digits=10, verbose_name='price per adult'),
        ),
```

Note that this field now has a default value of `Decimal('0.0')`.

I would expect that this field would behave like any other. That is, if no default value is specified and no explicit value is given, an error would be raised.

As this isn't happening, there's a risk of `tax_per_adult` silently being set to `0.0`.

I'm using django-money 0.13.2 and django 1.11.12.

GeeWee:
Fixes #411

Opening this PR now as I'm having trouble getting Tox to pass locally, and I'm hoping that Travis will give me more information as to why.


GeeWee:
Travis is failing as well, but master passed last time it was built.

This branch currently contains only an empty git commit, yet almost all environments fail.

Can anyone explain what I'm doing wrong?

benjaoming:
Oh wow, there's a new py-moneyed release out. I expect this is the explanation for why "123.45 EUR" has changed to "€123.45".

It mentions this in CHANGES.rst:

> DEFAULT locale is now used as a fallback to return a currency symbol if your chosen locale has no symbol set for that currency, rather than just returning the currency code.

benjaoming:
I suspect this commit did the change: https://github.com/limist/py-moneyed/commit/4fd08c8bfc6b1945d58f9f06c0a26130ae03a2a3

benjaoming:
So I'm not sure, but I think that this is intended behavior: If there is no locale defined, we should write `€123` and not `123 EUR` because this (and most other) particular ways of writing currencies which are listed [here](https://github.com/limist/py-moneyed/blob/4fd08c8bfc6b1945d58f9f06c0a26130ae03a2a3/moneyed/localization.py#L202) were put there with good reason and research.

@Stranger6667 any opinions?

Stranger6667:
In general, I believe that we shouldn't change `py-moneyed` behavior (maybe except for some missing features like rounding on Python 3) since the library intends to provide `py-moneyed` features for Django fields

aqeelvn:
@GeeWee 👋, wondering if you have any plan to work on this as the tests should not be failing anymore based on https://github.com/django-money/django-money/issues/451

GeeWee:
I do expect to take another crack at this, but it probably won't be in the next couple of weeks. If you want to have a go, feel free.

Stranger6667:
Fixed in #464

PR #453, opened 2018-12-23
Fix invalid Django 1.8 version check in `MoneyField.value_to_string`

Stranger6667:


djmoney/models/fields.py
```diff
@@ -286,7 +286,7 @@ def formfield(self, **kwargs):
         return super(MoneyField, self).formfield(**defaults)
 
     def value_to_string(self, obj):
-        if VERSION == (1, 8):
+        if VERSION[:2] == (1, 8):
```
benjaoming: Small tip, not necessary: But you can write this a bit like you would with pip: `Django>=1.8,<1.9`:

```
>>> (1, 8, 0) >= (1, 8) < (1, 9, 0)
True
```

benjaoming: So would be: `(1, 8) >= VERSION < (1,9)`

PR #457, opened 2019-01-06
Bump to 0.14.4

benjaoming:


Stranger6667:
The new pytest version removed `get_marker` and `get_closest_marker` should be used instead.

benjaoming:
Ah, it was an issue in pytest-cov 2.6.0, and pytest-cov 2.6.1 is now released, so simply rebuilding on Travis and it'll go away...

Stranger6667:
Hm, I didn't realize that it was in the coverage plugin :)
Btw, maybe we could consider this change to be a part of 0.14.4 release? https://github.com/django-money/django-money/pull/459/files

benjaoming:
Needed to rebuild w/o the cache, as 2.6.0 was still installed :man_shrugging: 

I have included #459 fixing #458 in the release :)

Issue #458, opened 2019-01-09
django rest framework get_value with amount=None and currency is not None throws an InvalidOperation exception


carvincarl:
In the DRF code, when deserializing a JSON object with a value for currency and None for the amount, there is an InvalidOperation exception instantiating the Money object.

Example stack trace:
```
.tox/django111-py36/lib/python3.6/site-packages/rest_framework/serializers.py:236: in is_valid
    self._validated_data = self.run_validation(self.initial_data)
.tox/django111-py36/lib/python3.6/site-packages/rest_framework/serializers.py:434: in run_validation
    value = self.to_internal_value(data)
.tox/django111-py36/lib/python3.6/site-packages/rest_framework/serializers.py:486: in to_internal_value
    primitive_value = field.get_value(data)
djmoney/contrib/django_rest_framework/fields.py:34: in get_value
    return Money(amount, currency)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'Money' object has no attribute 'amount'") raised in repr()] Money object at 0x10f4cb470>, amount = None
currency = 'USD'

    def __init__(self, amount=Decimal('0.0'), currency=DEFAULT_CURRENCY_CODE):
        if not isinstance(amount, Decimal):
>           amount = Decimal(str(amount))
E           decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]

.tox/django111-py36/lib/python3.6/site-packages/moneyed/classes.py:83: InvalidOperation
```

Here is the method with the problem:
```
    def get_value(self, data):
        amount = super(MoneyField, self).get_value(data)
        currency = data.get(get_currency_field_name(self.field_name), None)
        if currency:  # Missing a check for amount is None.
            return Money(amount, currency)
        return amount

```

The fix is to also check that the amount is not None.
```
        if currency and amount is not None:
            return Money(amount, currency)
```

Here is a test to expose the error:
```
    @pytest.mark.parametrize(
        'body, expected', (
            ({'field': '10', 'field_currency': 'EUR'}, Money(10, 'EUR')),
            ({'field': '12.20', 'field_currency': 'GBP'}, Money(12.20, 'GBP')),
[...]

PR #459, opened 2019-01-10
Fixed InvalidOperation in djmoney.contrib.django_rest_framework.fields.MoneyField.get_value

carvincarl:
Problem happens when amount is None and currency is not None.

Issue #458


benjaoming:
Thanks for fixing this and for completing the change log neatly!

PR #461, opened 2019-01-29
Added configurable decimal place

ZuluPro:
When I use the `Money` object manually, I'm not able to specify the number of decimals as it is hardcoded.

### DIFF
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -24,6 +24,10 @@ class Money(DefaultMoney):
     """
     use_l10n = None
 
+    def __init__(self, decimal_places=DECIMAL_PLACES, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


Stranger6667:
Nice change! However, `decimal_places` shouldn't be in the first place in `__init__` to be backward compatible. This one would work:

```
    def __init__(self, *args, **kwargs):
        self.decimal_places = kwargs.pop("decimal_places", DECIMAL_PLACES)
        super(Money, self).__init__(*args, **kwargs)
```

Or it may be better to specify arguments explicitly if you need to use `decimal_places` this way:

```
Money(100, "EUR", 5)
```

Stranger6667:
Closed via #477

PR #464, opened 2019-03-15
Remove implicit default value

washeck:
Since 189d8b65092a9e5b10c59e2b21a7a3dadfce4409 if no default
value is provided, value 0.0 and default currency is used.
This is opposite of what Django does and it makes it impossible
to specify not null fields with no default value.

Backwards incompatible change.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -189,9 +189,6 @@ class MoneyField(models.DecimalField):
         Field.creation_counter += 1
 
     def setup_default(self, default, default_currency, nullable):
-        if default is None and not nullable:
-            # Backwards compatible fix for non-nullable fields
-            default = 0.0
         if isinstance(default, string_types):
             try:
                 # handle scenario where default is formatted like:
@@ -208,7 +205,7 @@ class MoneyField(models.DecimalField):
             default = Money(default, default_currency)
         elif isinstance(default, OldMoney):
             default.__class__ = Money
-        if not (nullable and default is None) and not isinstance(default, Money):
+        if default is not None and not isinstance(default, Money):
             raise ValueError('default value must be an instance of Money, is: %s' % default)
         return default
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,15 @@
 Changelog
 =========
 
+`Unreleased`_
+----------------------
+
+Changed
+~~~~~~~
+
+- Remove implicit default value on non-nullable MoneyFields.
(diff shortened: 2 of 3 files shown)


washeck:
This attempts to fix #411 in backwards incompatible way.

I believe this is a step in the right direction but obviously the compatibility is an issue. If required, another possible approach would be to introduce a settings defaulting to the previous behavior but allowing to suppress the default value. I was thinking about something like USE_ZERO_AS_DEFAULT_AMOUNT. Then the modification in tests would not be needed.

I did not attempt at documentation change as I am not a native speaker and I am not sure if this PR will be accepted.

Stranger6667:
hej @washeck !
Thank you very much for your contribution :)
I like that it is explicit - less possibility to get unwanted behavior, especially with nullable fields

benjaoming:
We could fork `master` into a `0.14.x` release branch and merge this afterwards, provided that it gets a release note :)

@washeck thanks for the change and the motivation, I also agree with your reasoning. Would you be able to provide a release note, clarifying the backwards incompatibility?

washeck:
@benjaoming I added the release note.

Stranger6667:
What do you think about the type of the default value? 
Even if it is equal to `0.0` now, I think that it would be nice to encourage people to use `Decimal` or `int` instead of `float` as default values for `MoneyField`. How can we do it? Could having `0` as default in tests & examples be enough? Mb some runtime checks like in `py-moneyed` - if `float` is used, then a warning is raised?

washeck:
I agree that float should be discouraged (actually, the project I work on sets `decimal.getcontext().traps[decimal.FloatOperation] = 1` in app.ready).

On the other hand, I tried to keep the same behavior as django-money did before this change. I'm OK to change the default to `0` in this PR if you want.

Stranger6667:
> `decimal.getcontext().traps[decimal.FloatOperation] = 1`

This one is great :)

@benjaoming what do you think about types?

benjaoming:
Oh nice! Didn't know about that trap. Would it block this?

```python
MoneyField(max_digits=10, decimal_places=2, default=0.0, default_currency='USD')
```

And enforce this?

```python
MoneyField(max_digits=10, decimal_places=2, default=decimal.Decimal('0.00'), default_currency='USD')
```

?

I agree with the `FloatOperation` trap, but I don't like that the number of decimal places is specified twice in the above.. so we might wanna deprecate `decimal_places` in its place.

benjaoming:
Merging this into a new `0.15` release tracked in `master` and moved old `master` to `0.14.x`.

Maybe we could discuss the trap in a separate issue?

PR #467, opened 2019-03-18
Fix model-level validators behavior in REST Framework

rapIsKal:
now if moneyfield is done with validator like MinMoneyValidator(1) attempt to make a crud through drf fails because MinMoneyValidator is treated by drf as MinValueValidator. This patch fixes problem

### DIFF
--- a/djmoney/contrib/django_rest_framework/fields.py
+++ b/djmoney/contrib/django_rest_framework/fields.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
-from rest_framework.compat import MinValueValidator, MaxValueValidator
+from django.core.validators import MaxValueValidator, MinValueValidator
+
 from rest_framework.fields import empty
 from rest_framework.serializers import DecimalField, ModelSerializer
 
 from djmoney.models.fields import MoneyField as ModelField
-from djmoney.models.validators import MinMoneyValidator, MaxMoneyValidator
+from djmoney.models.validators import MaxMoneyValidator, MinMoneyValidator
 from djmoney.money import Money
 from djmoney.utils import MONEY_CLASSES, get_currency_field_name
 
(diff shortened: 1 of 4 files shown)


djmoney/contrib/django_rest_framework/fields.py
```diff
@@ -12,6 +14,18 @@ class MoneyField(DecimalField):
     does decimal's validation during transformation to native value.
     """
 
+    def __init__(self, *args, **kwargs):
+        # patches wrong validators for drf
+        super().__init__(*args, **kwargs)
+        validators = self.validators
+        print(type(validators))
```
Stranger6667: Could you pls, remove print?

djmoney/contrib/django_rest_framework/fields.py
```diff
@@ -12,6 +14,18 @@ class MoneyField(DecimalField):
...
+        for i in range(len(validators)):
+            if isinstance(validators[i], MinValueValidator):
+                validators[i] = MinMoneyValidator(self.min_value)
+            elif isinstance(validators[i], MaxValueValidator):
+                validators[i] = MaxMoneyValidator(self.max_value)
+        self.validators = validators
```
Stranger6667: Since `validators` is pointing to `self.validators` it shouldn't be required because the list is modified  during the `for` loop above

Stranger6667:
Hello @rapIsKal !
Thank you for your contribution! Could, you, pls:
- Add tests that will cover all cases for these validators
- Add a release note

Stranger6667:
The problem was a little bit deeper. By default, DRF takes only first instances of `MinValueValidator` and `MaxValueValidator` from `Field.validators`, and removes all other serializers of those types, which is not compatible with our approach to money validation, since we support validators for different currencies. With this change, all test cases are passing - I'll add a note to backward-incompatible changes list

Thank you for bringing this issue @rapIsKal

PR #470, opened 2019-03-27
Clarify how to use DRF serializer fields

washeck:
Update the docs to prevent confusion when user specifies only amount field on serializer.


washeck:
This one comes from real-world issue when we forgot to include the `*_currency` fields in serializer definition which lead to lengthy debugging of why the serialization does not work.

Stranger6667:
Thanks!

PR #472, opened 2019-04-02
fix deserializer so _get_model monkey patching is possible

Formulka:
Patching _get_model function is sometimes useful/necessary for migrations, good example is here: https://stackoverflow.com/a/5906258 if one needs to use a per-migration fixture. The import currently breaks this as it will always use the regular unpatched version of _get_model, simply moving it inside the Deserializer function fixes the issue.


Stranger6667:
Hello @Formulka !

Thank you for your contribution! Could you, please:
- Add a test case for the change (e.g. patching similar to the example you provided)
- A changelog entry
- Rebase against current master

Issue #475, opened 2019-04-13
Add default currency for DRF field


Stranger6667:
Could be helpful - you don't have to specify currency everytime

PR #479, opened 2019-04-20
Add default currency for DRF field #475

butorov:
What do you think? :)
If the approach is ok, I may add some description into the documentation.

### DIFF
--- a/tests/contrib/test_django_rest_framework.py
+++ b/tests/contrib/test_django_rest_framework.py
@@ -3,6 +3,8 @@ from decimal import Decimal
 
 import pytest
 
+import six
+from djmoney.contrib.django_rest_framework import MoneyField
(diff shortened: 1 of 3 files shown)


tests/contrib/test_django_rest_framework.py
```diff
@@ -3,6 +3,8 @@
 
 import pytest
 
+import six
+from djmoney.contrib.django_rest_framework import MoneyField
```
Stranger6667: Could you make this import local to avoid failing test build?

Stranger6667:
Yep, I like it, could you, pls, fix the note above? then we can merge it

Stranger6667:
Could you rebase, please?

butorov:
Sure! Done.

PR #482, opened 2019-04-20
Support for customization of ``CurrencyField`` length. Closes #480

### LINKED ISSUE #480: Allow changing max_length in CurrencyField (cryptocurrency support)
Now CurrencyField overrides max_length to 3, which is true for currencies but not for cryprocurrencies, for example USDT.
I'm not sure, but I guess simply changing
```
def __init__(self, price_field=None, default=DEFAULT_CURRENCY, **kwargs):
    if isinstance(default, Currency):
        default = default.code
    kwargs['max_length'] = 3
    self.price_field = price_field
    super(CurrencyField, self).__init__(default=default, **kwargs)
```
to 
```
def __init__(self, price_field=None, default=DEFAULT_CURRENCY, **kwargs):
    if isinstance(default, Currency):
        default = default.code
    if 'max_length' not in kwargs:
        kwargs['max_length'] = 3
    self.price_field = price_field
    super(CurrencyField, self).__init__(default=default, **kwargs)
```
will do the trick.

I would be extremely thankful for that.

Stranger6667:
Some cryptocurrencies could have codes longer than three characters

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -157,7 +157,7 @@ class CurrencyField(models.CharField):
     def __init__(self, price_field=None, default=DEFAULT_CURRENCY, **kwargs):
         if isinstance(default, Currency):
             default = default.code
-        kwargs['max_length'] = 3
+        kwargs.setdefault("max_length", 3)
         self.price_field = price_field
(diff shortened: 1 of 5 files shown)


Stranger6667:
Thank you, Ben :)

PR #485, opened 2019-04-22
Black

Stranger6667:
It could look like this

### DIFF
--- a/djmoney/__init__.py
+++ b/djmoney/__init__.py
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
 
-__version__ = '0.15.dev'
-default_app_config = 'djmoney.apps.MoneyConfig'
+__version__ = "0.15.dev"
+default_app_config = "djmoney.apps.MoneyConfig"
--- a/tox.ini
+++ b/tox.ini
@@ -30,15 +30,19 @@ usedevelop = false
 deps =
 	flake8
 	isort
+    black
 commands =
     flake8 djmoney tests
     isort -rc -c {toxinidir}/djmoney {toxinidir}/tests
+    black -l 120 --check {toxinidir}/djmoney {toxinidir}/tests
 
 [testenv:fix-lint]
 deps =
 	isort
+    black
 commands =
     isort -rc {toxinidir}/djmoney {toxinidir}/tests
+    black -l 120 {toxinidir}/djmoney {toxinidir}/tests
 
 [django]
 1.8.x  =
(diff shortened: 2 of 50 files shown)


djmoney/__init__.py
```diff
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
 
-__version__ = '0.15.dev'
-default_app_config = 'djmoney.apps.MoneyConfig'
+__version__ = "0.15.dev"
+default_app_config = "djmoney.apps.MoneyConfig"
```
benjaoming: Can black be made to not care about quotes? It's going to be annoying :)

washeck: See https://github.com/python/black#strings

The whole point of Black is about not being configurable.

Stranger6667: Nope, there is only one option :) It is almost uncompromising :) There were some good arguments for always using double quotes in one issue in black, but I don't remember where exactly. Usually, I set up a pre-commit hook for black and everything works without hassle

Stranger6667: aha, it is exactly what I was thinking about :)

benjaoming: I (almost) always used double quotes for variable/human strings and single quotes for keys and other sorts of constant-ish strings :)

benjaoming:
So TBH I don't see Black solving problems that I experienced.. we already had linting.. it provides a very huge changeset that can be annoying for backporting purposes. But I don't mind it, so if you or anyone else find that the code is more readable or the linting process has become simplified, then that's fine :)

Stranger6667:
Hej @benjaoming !

What problems did you experience? Maybe we can tune our linting somehow to fix/prevent them?

*Re Black*: I think, that the most beneficial thing is not having any formatting discussions anymore (not sure if we had any in django-money, though). Also, the backporting process is quite straightforward - I usually accept changes from the target branch and apply `black` again. From my experience - at my previous job, we usually had a couple of comments about formatting and now at the current job, we don't have them at all which actually reduces the review time. 

Regards,
Dmitry

benjaoming:
> What problems did you experience?

None :)

Sounds good that Black can be re-applied for solving merge-conflicts, that's a good plan.

tox.ini
```diff
@@ -30,15 +30,19 @@ usedevelop = false
 deps =
 	flake8
 	isort
+    black
 commands =
     flake8 djmoney tests
     isort -rc -c {toxinidir}/djmoney {toxinidir}/tests
+    black -l 120 --check {toxinidir}/djmoney {toxinidir}/tests
```
benjaoming: does it replace the existing commands btw?

PR #486, opened 2019-04-23
Remove default=None from deconstruct

washeck:
Deconstruct should include only kwargs which are different from
defaults. Without this fix, running makemigrations after #464
was merged results in AlterField operation with default=None
which is not needed as it is already the default value.


Stranger6667:
Thanks! :)

Issue #490, opened 2019-05-14
Confusion in default_currency type


washeck:
There seems to be a confusion in how we understand default currency. The README says you can set it to None like this
`MoneyField(max_digits=10, decimal_places=2, null=True, default_currency=None)`

Unfortunately, this does not work for me because `MoneyField.deconstruct()` has the following code:
```
        if self.default_currency != DEFAULT_CURRENCY:
            kwargs['default_currency'] = str(self.default_currency)
```
which leads to funny situation when you end up with string 'None' in the db.

DEFAULT_CURRENCY is not documented in README but there is a comment in `settings.py`
```
# The default currency, you can define this in your project's settings module
# This has to be a currency object imported from moneyed
DEFAULT_CURRENCY = getattr(settings, 'DEFAULT_CURRENCY', DEFAULT_CURRENCY)
```

I find it confusing to have DEFAULT_CURRENCY in settings as object and default_currency in kwargs as string. I can submit PR but I am not sure what approach you prefer. As a minimal fix I suggest to special-case None in the kwargs and to add asserts for expected types (str or Currency).

benjaoming:
Sounds like a bug. @Stranger6667 should we implement a fix that follows the current docs? Seems least dramatic?

Thinking something like:

```python
        if self.default_currency and self.default_currency != DEFAULT_CURRENCY:
            kwargs['default_currency'] = str(self.default_currency)
```

Stranger6667:
Yep, it is a bug - let's start with a simple fix. There are some confusing points that we can clarify/eliminate as the next steps to `1.0` release.

1. `default` vs `default_currency` #474 ;
2. Possibility to use `float` as the default value (I think we need to forbid it);
3. Multiple formats for the `MoneyField`. In general, I think that we need to be stricter. Now we have - `"10 EUR"`, `(10, "EUR")`, `Money(10, "EUR")` or numerics;
4. `moneyed.Money` vs `djmoney.money.Money`, the first one doesn't support `F` expressions + some arithmetics and exchange conversions. Also, there is `MoneyPatched` compatibility class that is deprecated;

We have some extra logic to validate the length of the passed tuple, string, etc. And I'm not sure if there is a real use case for strings and tuples/lists as default values.

I think it would be nice to have some restrictions on our side, that would prevent possible bugs or precision loss on the user side - `float` defaults for example.

washeck:
@Stranger6667 Absolutely agreed with your points. 

[...]

PR #491, opened 2019-05-16
Assert that it's possible to have DEFAULT_CURRENCY=None and a nullable MoneyField

benjaoming:
#490

CC: @washeck - would this work?

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -295,7 +295,7 @@ class MoneyField(models.DecimalField):
             del kwargs["default"]
         else:
             kwargs["default"] = self.default.amount
-        if self.default_currency != DEFAULT_CURRENCY:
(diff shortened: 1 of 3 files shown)


washeck:
@benjaoming Yes, this looks perfect to me. Thank you.

Stranger6667:
Sorry, I didn't see that it is against `0.14.x` :( Is it intended to be like this?

@benjaoming

benjaoming:
> Is it intended to be like this?

Yeah, I wanted it like this because it behaves like a bug fix.. otherwise, it'd have to wait for 1.0 :)

benjaoming:
Ah okay, let's wait until 1.0 then and not kick around this older branch :+1:

benjaoming:
Added a release note

Issue #488, opened 2019-05-07
MoneyField with manually linked CurrencyField


briancappello:
Hi,

I'm working on a project where the concept of `currency` is tied to a model, and all the `MoneyField` amounts on that model (or related models) should get their currency from the single `currency` field. For example:

```python
from django.db import models

class Listing(models.Model):
    currency = CurrencyField()
    reserve = MoneyField(currency, ...)
    # bids -> [Bid]

class Bid(models.Model):
    listing = models.ForeignKey(Listing, related_name='bids', on_delete=models.CASCADE)
    amount = MoneyField('listing__currency', ...)
```

In order to support this, I ended up having to make some modifications to `MoneyField` and `MoneyFieldProxy`. I'm not sure if I'm just misunderstanding something, or if this is actually an opportunity for a PR? Possibly related to issue #487. Example implementation follows:

```python
from django.db.models import Func, Value
from django.db.models.expressions import BaseExpression
from djmoney.money import Money
from djmoney.models.fields import (
    MoneyField as BaseMoneyField,
    MoneyFieldProxy as BaseMoneyFieldProxy,
    get_currency_field_name,
    prepare_expression,
    validate_money_expression,
)
from djmoney.settings import CURRENCY_CHOICES, DEFAULT_CURRENCY


class MoneyFieldProxy(BaseMoneyFieldProxy):
    """
    Overridden to support fetching currency across relationships.
    """

    def __init__(self, field, currency_field_name=None):
        self.field = field
        self.currency_field_name = currency_field_name or get_currency_field_name(
            self.field.name, self.field
        )

    def _money_from_obj(self, obj):
        amount = obj.__dict__[self.field.name]
        if "__" in self.currency_field_name:
            rel, name = self.currency_field_name.split("__")
            related = getattr(obj, rel)
            currency = getattr(related, name)
        else:
            currency = obj.__dict__[self.currency_field_name]
        if amount is None:
            return None
        return Money(amount=amount, currency=currency)

    def __set__(self, obj, value):  # noqa
        if isinstance(value, BaseExpression):
            if isinstance(value, Value):
                value = self.prepare_value(obj, value.value)
            elif not isinstance(value, Func):
                validate_money_expression(obj, value)
                prepare_expression(value)
        else:
            value = self.prepare_value(obj, value)
        obj.__dict__[self.field.name] = value


def noop_add_currency_field(*args, **kwargs):
    pass


class LinkedCurrencyMoneyField(BaseMoneyField):
    """
[...]

PR #492, opened 2019-05-17
add LinkedCurrencyMoneyField

briancappello:
Implements https://github.com/django-money/django-money/issues/488

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -141,6 +141,42 @@ class MoneyFieldProxy(object):
             setattr(obj, self.currency_field_name, value)
 
 
+class LinkedCurrencyMoneyFieldProxy(MoneyFieldProxy):
+    """
+    Overridden to support fetching currency across relationships.
+    """
+
+    def __init__(self, field, currency_field_name=None):
+        super().__init__(field)
+        self.currency_field_name = currency_field_name or get_currency_field_name(
+            self.field.name, self.field
+        )
+
+    def _money_from_obj(self, obj):
+        amount = obj.__dict__[self.field.name]
+        if "__" in self.currency_field_name:
+            *related_attrs, currency_attr = self.currency_field_name.split("__")
+            for related_attr in related_attrs:
+                obj = getattr(obj, related_attr)
+            currency = getattr(obj, currency_attr)
+        else:
+            currency = obj.__dict__[self.currency_field_name]
+        if amount is None:
+            return None
+        return Money(amount=amount, currency=currency)
+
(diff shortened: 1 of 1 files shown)


djmoney/models/fields.py
```diff
@@ -141,6 +141,42 @@ def set_currency(self, obj, value):
             setattr(obj, self.currency_field_name, value)
 
 
+class LinkedCurrencyMoneyFieldProxy(MoneyFieldProxy):
+    """
+    Overridden to support fetching currency across relationships.
+    """
+
+    def __init__(self, field, currency_field_name=None):
+        super().__init__(field)
```
Stranger6667: We still support Python 2.7, where this syntax is invalid

briancappello: Ah gotchya, ok, lemme fix that (there are a few other 3.x only lines of code too)

briancappello: @Stranger6667 2.7 compat fixes pushed

benjaoming:
Fixed a single line of non-linting code, I love black, this only delayed the PR by under 1 year :)

benjaoming:
Why can't this behavior be added to MoneyField rather than adding a new `LinkedCurrencyMoneyField` ?

Would that produce migrations for everyone using django-money and upgrading or can it silently be added as a new kwarg for `MoneyField` ?

benjaoming (changes requested):
Sorry about the staleness.

I don't have any comments on the implementation, I think it's great that it seems to not generate migrations on existing implementations.

But it should really have a test or two.

RmaxTwice:
Hello guys, this PR is something I would like to use in a current project I'm working on, what would it require to be merged? just some tests?

Stranger6667:
Hi @RmaxTwice 

Yes, I'd be happy to see tests for querying a model with this field and see how it works with migrations. Maybe a little revert of changes relevant to my comment above, as we dropped support for Python 2.7.

RmaxTwice:
Could you please give me a brief explanation on what the `fields.py/MoneyFieldProxy` class does? because I don't see an explicit use for it but noticed that @benjaoming created an extended class of it. Just so maybe I can help to get this PR going since it covers a very useful use case.

benjaoming:
@RmaxTwice a contribution with tests of this would be greatly appreciated! Maybe in that process, it can be uncovered and documented/commented what `MoneyFieldProxy` is for, I can't recall it.

benjaoming:
This is a super-interesting feature and I wish it was ready! Doesn't feel good to close the PR, but is anyone in here around to continue the work?

PR #501, opened 2019-06-19
Getting a Money object with decimal places from a MoneyField

astutejoe:
Right now when getting a MoneyField it returns a Money object without the MoneyField decimal places, producing unexpected behaviour when printing/converting to a string. I'm not sure if that works as I don't have the time to test so consider this as an issue with a probable pull request. It's my first pull request so sorry about anything, and have a nice day :)

image

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -93,7 +93,7 @@ class MoneyFieldProxy(object):
         currency = obj.__dict__[self.currency_field_name]
         if amount is None:
             return None
-        return Money(amount=amount, currency=currency)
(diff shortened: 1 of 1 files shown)


astutejoe:
Tested now out of euphoria :D
Might this be my first pull request? ;D
image

benjaoming:
This makes sense. As you remark, it seems important since some displays of prices/values may show wrongly.

@Stranger6667 this would make sense as a 0.15.1 release?

benjaoming:
Great work @astutejoe - welcome as a contributor :)

PR #502, opened 2019-06-22
0.15.1

benjaoming:
CC: @astutejoe


benjaoming:
This has been released as of today. Sorry that I wasn't faster at following up on it.

Issue #514, opened 2019-10-14
Add Django 3 support, remove Python 2.7 support


Hispar:
Better to add `six` as a dependency for compatibility.

https://docs.djangoproject.com/en/dev/releases/3.0/#removed-private-python-2-compatibility-apis

benjaoming:
Thanks for reporting @Hispar - I would propose that we remove Python 2.7 support entirely, so we don't complicate the app any further.

People depending on 2.7 can use older releases.

@Stranger6667 does this sound like an okay plan for a 0.16 release?

Stranger6667:
It sounds great :)

benjaoming:
Just correcting myself here: The next release is not 0.16, but in fact the 1.0 we have been talking about for so long :)

benjaoming:
Fixed in #515

@Stranger6667 looking at the 1.0 milestone, we have just 1 issue open: https://github.com/django-money/django-money/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.0

PR #515, opened 2019-10-14
Add Django 3.0 support, remove Python 2.7 support

benjaoming:
#514 

TODO:

 * [ ] Clean up unicode-related stuff from Python 2.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,17 +30,15 @@ matrix:
       env: TOX_ENV=django21-py37
     - python: 3.6
       env: TOX_ENV=django21-py36
+    - python: 3.6
+      env: TOX_ENV=django111-py36
     - python: 3.5
       env: TOX_ENV=django21-py35
     - python: pypy3
       env: TOX_ENV=django21-pypy3
-    - python: 3.6
-      env: TOX_ENV=django111-py36
     - python: 3.5
       env: TOX_ENV=django111-py35
-    - python: 2.7
-      env: TOX_ENV=django111-py27
-    - python: pypy2.7-5.8.0
-      env: TOX_ENV=django111-pypy
+    - python: 3.5
+      env: TOX_ENV=django22-py35
 after_success:
   - codecov
--- a/djmoney/models/managers.py
+++ b/djmoney/models/managers.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
+from functools import wraps
+
 from django.db.models import Case, F, Q
 from django.db.models.constants import LOOKUP_SEP
 from django.db.models.expressions import BaseExpression
 from django.db.models.fields import FieldDoesNotExist
-from django.utils.six import wraps
 
 from .._compat import smart_unicode
 from ..utils import MONEY_CLASSES, get_currency_field_name, prepare_expression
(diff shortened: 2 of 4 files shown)


lundberg:
Lovely PR!

FYI, tests on travis are passing *(except linting)*, but when trying this PR locally in a project, I get an `ImportError` referring to `djmoney.serializers` importing `django.utils.six` [here](https://github.com/benjaoming/django-money/blob/6501a2f8383e1e3307846190159587527c847b56/djmoney/serializers.py#L6).

Not covered?

benjaoming:
Thanks @lundberg :) This should be working now.

Moreover, the `flake8` command wasn't working in Travis CI.. so important stuff such as undeclared variables was missing in the linting.

@Stranger6667 would you be able to take a look at this and merge at will?

djmoney/serializers.py
```diff
@@ -25,7 +23,7 @@ def Deserializer(stream_or_string, **options):  # noqa
 
     ignore = options.pop("ignorenonexistent", False)
 
-    if not isinstance(stream_or_string, (bytes, six.string_types)):
```
Stranger6667: It seemed we had a bug - `string_types` is a tuple, which makes the check incorrect :) 
Currently on Python 3:

```
(bytes, (str,))
```

but after the change it is correct :)

djmoney/_compat.py
```diff
@@ -1,33 +1,10 @@
...
-    from importlib import reload as reload_module
-
-try:
-    from urllib2 import urlopen
-    from urlparse import urlparse, parse_qsl, urlunparse
-except ImportError:
-    from urllib.request import urlopen
-    from urllib.parse import urlparse, parse_qsl, urlunparse
+# Formerly Python 2.7 diverted, so stored here
+string_types = (str, bytes)
```
Stranger6667: Do we need it anymore? It seems like we can inline all this code since there is no extra difference in Python versions and `_compat` is not that needed ... What do you think?

benjaoming: This was me being sloppy :) But also kind of wanting to keep `_compat` because of the concept: Isolating this compatibility in places where it can easily be removed.

Not removing it despite of that confirmed maybe the extreme stubborn laziness of a programmer.. "oh a corner, let me cut that" :D

djmoney/serializers.py
benjaoming: Oh yeah you're right, haha. But somehow it still works.. I can't explain:

    >>> isinstance("test", (bytes, (str,),))
    True

djmoney/_compat.py
Stranger6667: :)))

djmoney/serializers.py
Stranger6667: Hm, it is strange :) I didn't know that it works this way. It is working even for more nesting :))

```
>>> isinstance(1.0, (bytes, (int, (float, ))))
True
```

benjaoming:
@Stranger6667 - the last change was necessary, since we had specified `universal=True` for building wheels, which says that the .whl will work on both Python 2 and 3.

Now, we just create .whl for Python 3, anyone running `pip install django-money` on Python 2 will get version 0.15, and Python 3 environments will get version 1.0 (when it's released).

PR #510, opened 2019-09-23
Fixed currency_widget handling.

atleta:
I've made a small fix: the `currency_widget` parameter of `forms.fields.MoneyField.__init__` was handled incorrectly when passed in a (non-None) value. It got assigned to the self.widget (i.e. used as the widget for the MoneyField) istead of the being used as the currency widget of  the MoneyWidget.

PR #518, opened 2019-10-16
Fixed currency_widget handling.

atleta:
I've made a small fix: the currency_widget parameter of forms.fields.MoneyField.__init__ was handled incorrectly when passed in a (non-None) value. It got assigned to the self.widget (i.e. used as the widget for the MoneyField) istead of the being used as the currency widget of the MoneyWidget.


atleta:
@benjaoming The content is the same as in #510 , I've rebased my commit onto the current master head.

Issue #99, opened 2014-11-06
Magic money querying ability does not chain


spookylukey:
Take, for example, `VanillaMoneyFieldTestCase.testCurrencySearch`. In the line

    shouldBeEmpty = ModelWithVanillaMoneyField.objects.filter(money__lt=otherMoney)

if you replace `objects` with `objects.all().all()`, then the test fails.

Basically, you end up with a vanilla `QuerySet` that does not 'understand money'. However, there is no warning - the query appears to succeed, but returns the wrong answer. This hugely violates expectations - people normally chain ORM methods a lot, and don't expect them to suddenly behave differently because of an earlier `filter` or `order_by` etc.

This is a serious flaw in this feature, IMO. If it can't be fixed, I'd recommend pulling it completely. The alternative method (explicit filtering against the `_currency` field) will always work.

To solve it, I think you need a custom `QuerySet` subclass. It is this that should have `understands_money` applied. And you then need to do patching that will return this `QuerySet` rather than anything else. I may be able to work something up for this.

Stranger6667:
@benjaoming , I think it is not relevant anymore - chaining works fine.

benjaoming:
Closing :)

PR #519, opened 2019-10-16
Fix order_by() not being wrapped by understand_money() 

lieryan:
This is probably related to https://github.com/django-money/django-money/issues/99, but
money_manager didn't work correctly with order_by. This fails with both custom
money_manager() and the automatically generated manager.

# Minimal test case

## tests.py
   
    from django.test import TestCase
   
    from chaining.models import Question
   
   
    class QuestionTest(TestCase):
        def test_prize(self):
            question_aud = Question.objects.create(prize=Money(10, 'AUD'))
            question_usd = Question.objects.create(prize=Money(10, 'USD'))
   
            # passes as expected
            self.assertQuerysetEqual(
                Question.objects.filter(prize=Money(10, 'AUD')),
                ['<Question: Q#1 - A$10.00>'],
            )
   
            # however, order_by didn't work correctly
            self.assertQuerysetEqual(
                Question.objects.order_by('id').filter(prize=Money(10, 'AUD')),
                ['<Question: Q#1 - A$10.00>'],
                # actual value: ['<Question: Q#1 - A$10.00>', '<Question: Q#2 - $10.00>']
            )

## models.py

    from django.db import models
    from djmoney.models.fields import MoneyField
   
   
    class Question(models.Model):
        prize = MoneyField(decimal_places=2, max_digits=10)
   
        def __str__(self):
            return 'Q#{} - {}'.format(self.pk, self.prize)

### DIFF
--- a/djmoney/models/managers.py
+++ b/djmoney/models/managers.py
@@ -206,7 +206,7 @@ def understands_money(func):
     return wrapper
 
 
-RELEVANT_QUERYSET_METHODS = ("distinct", "get", "get_or_create", "filter", "exclude", "update")
+RELEVANT_QUERYSET_METHODS = ("distinct", "get", "get_or_create", "filter", "exclude", "update", "order_by")
 EXPAND_EXCLUSIONS = {"get_or_create": ("defaults",)}
 
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -10,7 +10,7 @@ Added
 ~~~~~
 
 - Support for money descriptor customization. (`Stranger6667`_)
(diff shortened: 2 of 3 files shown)


benjaoming:
Thanks so much @lieryan - this fix can go into the 1.0 release.

Would you be able to add a test case in `tests/` and a release note in `docs/changes.rst`?

lieryan:
Hi @benjaoming, thank you for the prompt response. I've updated the pull request to include the requested changes.

benjaoming:
Perfect PR, @lieryan :100: 

Thanks, will merge and it will be included in 1.0

PR #523, opened 2019-10-27
Applied some more Python 2 cleanup

claudep:


benjaoming:
Agreed, these are really nice changes, thanks @claudep

Issue #509, opened 2019-09-18
Error with decimal_places equals to zero


arnaudlimbourg:
A model using a ``MoneyField(max_digits=2, decimal_places=0)`` used in the admin started giving errors after upgrading to 0.15

After investigating the best explanation we came up with is the introduction of the extra ``
 "decimal_places": DECIMAL_PLACES`` (see https://github.com/django-money/django-money/blob/master/djmoney/models/fields.py#L280)

We had no ``DECIMAL_PLACES`` defined in our configuration so the default value of 2 was used and not the ``decimal_places=0`` definition on the field. It was impossible to edit the model object in the admin (though it did work in shell).

Setting a ``DECIMAL_PLACES=0`` makes it work but seems to be the wrong fix.

Not knowing how the internals of django admin I'm not sure what django money should do. it seems that something like the following code is the idea of what should happen.

```python
defaults = {"form_class": forms.MoneyField, "decimal_places": self.decimal_places or DECIMAL_PLACES}
```

Has anyone seen errors of that kind?

jglstewart:
I am having a very similar issue. I am getting the following from admin validation:

> Ensure that there are no more than 2 decimal places

Even though, my model looks like this:

`demand = MoneyField(max_digits=5, decimal_places=4, default_currency='CAD',blank=True, null=True)
`

When I add

> CURRENCY_DECIMAL_PLACES = 0

to my settings file, the error changes to 0 decimal places.

Why bother having a default setting if the model field requires a decimal_places attribute?

Furthermore, why is it not using the model field setting?

`>> RateCodeCharge._meta.get_field('demand').decimal_places`


`>> 4`

benjaoming:
Good point @arnaudlimbourg - thanks for filing! That would make for a welcome PR!

PR #524, opened 2019-11-05
Set default decimal_places value to settings configuration and use de…

fara:
By reading @arnaudlimbourg I created this PR to fix issue #509 
Did a slight change since his proposed solution would not work when decimal_places is 0 (evaluate to False)


benjaoming:
Dear @fara - thanks for the PR fix. There is a failing test, would you be able to investigate this?

fara:
Done @benjaoming 
Added a new test to assert the order of precendence

PR #532, opened 2019-11-27
WIP: Fix #530 -- Keep None as `default_currency` in `deconstruct()`

### LINKED ISSUE #530: `null=True, default_currency=None` on added/altered fields does not set null properly
I've added a Money field with null=True and default_currency=Non to an existing model, but it results in a SQL migration that does not make the currency field nullable. The python migration does, but it doesn't translate correctly.

```python
from django.db import migrations
import djmoney.models.fields


class Migration(migrations.Migration):

    dependencies = [
        ('core', '0001_initial'),
    ]

    operations = [
        migrations.AddField(
            model_name='examplemodelwithlateraddedmoneyfield',
            name='display',
            field=djmoney.models.fields.MoneyField(decimal_places=4, max_digits=20, null=True),
        ),
        migrations.AddField(
            model_name='examplemodelwithlateraddedmoneyfield',
            name='display_currency',
            field=djmoney.models.fields.CurrencyField(choices=[('XUA', 'ADB Unit of Account'), ('AFN', 'Afghani'), ('DZD', 'Algerian Dinar'), ('ARS', 'Argentine Peso'), ('AMD', 'Armenian Dram'), ('AWG', 'Aruban Guilder'), ('AUD', 'Australian Dollar'), ('AZN', 'Azerbaijanian Manat'), ('BSD', 'Bahamian Dollar'), ('BHD', 'Bahraini Dinar'), ('THB', 'Baht'), ('PAB', 'Balboa'), ('BBD', 'Barbados Dollar'), ('BYN', 'Belarussian Ruble'), ('BYR', 'Belarussian Ruble'), ('BZD', 'Belize Dollar'), ('BMD', 'Bermudian Dollar (customarily known as Bermuda Dollar)'), ('BTN', 'Bhutanese ngultrum'), ('VEF', 'Bolivar Fuerte'), ('BOB', 'Boliviano'), ('XBA', 'Bond Markets Units European Composite Unit (EURCO)'), ('BRL', 'Brazilian Real'), ('BND', 'Brunei Dollar'), ('BGN', 'Bulgarian Lev'), ('BIF', 'Burundi Franc'), ('XOF', 'CFA Franc BCEAO'), ('XAF', 'CFA franc BEAC'), ('XPF', 'CFP Franc'), ('CAD', 'Canadian Dollar'), ('CVE', 'Cape Verde Escudo'), ('KYD', 'Cayman Islands Dollar'), ('CLP', 'Chilean peso'), ('XTS', 'Codes specifically reserved for testing purposes'), ('COP', 'Colombian peso'), ('KMF', 'Comoro Franc'), ('CDF', 'Congolese franc'), ('BAM', 'Convertible Marks'), ('NIO', 'Cordoba Oro'), ('CRC', 'Costa Rican Colon'), ('HRK', 'Croatian Kuna'), ('CUP', 'Cuban Peso'), ('CUC', 'Cuban convertible peso'), ('CZK', 'Czech Koruna'), ('GMD', 'Dalasi'), ('DKK', 'Danish Krone'), ('MKD', 'Denar'), ('DJF', 'Djibouti Franc'), ('STD', 'Dobra'), ('DOP', 'Dominican Peso'), ('VND', 'Dong'), ('XCD', 'East Caribbean Dollar'), ('EGP', 'Egyptian Pound'), ('SVC', 'El Salvador Colon'), ('ETB', 'Ethiopian Birr'), ('EUR', 'Euro'), ('XBB', 'European Monetary Unit (E.M.U.-6)'), ('XBD', 'European Unit of Account 17(E.U.A.-17)'), ('XBC', 'European Unit of Account 9(E.U.A.-9)'), ('FKP', 'Falkland Islands Pound'), ('FJD', 'Fiji Dollar'), ('HUF', 'Forint'), ('GHS', 'Ghana Cedi'), ('GIP', 'Gibraltar Pound'), ('XAU', 'Gold'), ('XFO', 'Gold-Franc'), ('PYG', 'Guarani'), ('GNF', 'Guinea Franc'), ('GYD', 'Guyana Dollar'), ('HTG', 'Haitian gourde'), ('HKD', 'Hong Kong Dollar'), ('UAH', 'Hryvnia'), ('ISK', 'Iceland Krona'), ('INR', 'Indian Rupee'), ('IRR', 'Iranian Rial'), ('IQD', 'Iraqi Dinar'), ('IMP', 'Isle of Man Pound'), ('JMD', 'Jamaican Dollar'), ('JOD', 'Jordanian Dinar'), ('KES', 'Kenyan Shilling'), ('PGK', 'Kina'), ('LAK', 'Kip'), ('KWD', 'Kuwaiti Dinar'), ('AOA', 'Kwanza'), ('MMK', 'Kyat'), ('GEL', 'Lari'), ('LVL', 'Latvian Lats'), ('LBP', 'Lebanese Pound'), ('ALL', 'Lek'), ('HNL', 'Lempira'), ('SLL', 'Leone'), ('LSL', 'Lesotho loti'), ('LRD', 'Liberian Dollar'), ('LYD', 'Libyan Dinar'), ('SZL', 'Lilangeni'), ('LTL', 'Lithuanian Litas'), ('MGA', 'Malagasy Ariary'), ('MWK', 'Malawian Kwacha'), ('MYR', 'Malaysian Ringgit'), ('TMM', 'Manat'), ('MUR', 'Mauritius Rupee'), ('MZN', 'Metical'), ('MXV', 'Mexican Unidad de Inversion (UDI)'), ('MXN', 'Mexican peso'), ('MDL', 'Moldovan Leu'), ('MAD', 'Moroccan Dirham'), ('BOV', 'Mvdol'), ('NGN', 'Naira'), ('ERN', 'Nakfa'), ('NAD', 'Namibian Dollar'), ('NPR', 'Nepalese Rupee'), ('ANG', 'Netherlands Antillian Guilder'), ('ILS', 'New Israeli Sheqel'), ('RON', 'New Leu'), ('TWD', 'New Taiwan Dollar'), ('NZD', 'New Zealand Dollar'), ('KPW', 'North Korean Won'), ('NOK', 'Norwegian Krone'), ('PEN', 'Nuevo Sol'), ('MRO', 'Ouguiya'), ('TOP', 'Paanga'), ('PKR', 'Pakistan Rupee'), ('XPD', 'Palladium'), ('MOP', 'Pataca'), ('PHP', 'Philippine Peso'), ('XPT', 'Platinum'), ('GBP', 'Pound Sterling'), ('BWP', 'Pula'), ('QAR', 'Qatari Rial'), ('GTQ', 'Quetzal'), ('ZAR', 'Rand'), ('OMR', 'Rial Omani'), ('KHR', 'Riel'), ('MVR', 'Rufiyaa'), ('IDR', 'Rupiah'), ('RUB', 'Russian Ruble'), ('RWF', 'Rwanda Franc'), ('XDR', 'SDR'), ('SHP', 'Saint Helena Pound'), ('SAR', 'Saudi Riyal'), ('RSD', 'Serbian Dinar'), ('SCR', 'Seychelles Rupee'), ('XAG', 'Silver'), ('SGD', 'Singapore Dollar'), ('SBD', 'Solomon Islands Dollar'), ('KGS', 'Som'), ('SOS', 'Somali Shilling'), ('TJS', 'Somoni'), ('SSP', 'South Sudanese Pound'), ('LKR', 'Sri Lanka Rupee'), ('XSU', 'Sucre'), ('SDG', 'Sudanese Pound'), ('SRD', 'Surinam Dollar'), ('SEK', 'Swedish Krona'), ('CHF', 'Swiss Franc'), ('SYP', 'Syrian Pound'), ('BDT', 'Taka'), ('WST', 'Tala'), ('TZS', 'Tanzanian Shilling'), ('KZT', 'Tenge'), ('XXX', 'The codes assigned for transactions where no currency is involved'), ('TTD', 'Trinidad and Tobago Dollar'), ('MNT', 'Tugrik'), ('TND', 'Tunisian Dinar'), ('TRY', 'Turkish Lira'), ('TMT', 'Turkmenistan New Manat'), ('TVD', 'Tuvalu dollar'), ('AED', 'UAE Dirham'), ('XFU', 'UIC-Franc'), ('USD', 'US Dollar'), ('USN', 'US Dollar (Next day)'), ('UGX', 'Uganda Shilling'), ('CLF', 'Unidad de Fomento'), ('COU', 'Unidad de Valor Real'), ('UYI', 'Uruguay Peso en Unidades Indexadas (URUIURUI)'), ('UYU', 'Uruguayan peso'), ('UZS', 'Uzbekistan Sum'), ('VUV', 'Vatu'), ('CHE', 'WIR Euro'), ('CHW', 'WIR Franc'), ('KRW', 'Won'), ('YER', 'Yemeni Rial'), ('JPY', 'Yen'), ('CNY', 'Yuan Renminbi'), ('ZMK', 'Zambian Kwacha'), ('ZMW', 'Zambian Kwacha'), ('ZWD', 'Zimbabwe Dollar A/06'), ('ZWN', 'Zimbabwe dollar A/08'), ('ZWL', 'Zimbabwe dollar A/09'), ('PLN', 'Zloty')],
				default=None, editable=False, max_length=3, null=True),
        ),
    ]
```
The resulting SQL migration looks like the following:
```sql
BEGIN;                                                                             
--
-- Add field display to examplemodelwithlateraddedmoneyfield
--
ALTER TABLE "core_examplemodelwithlateraddedmoneyfield" ADD COLUMN "display" numeric(20, 4) NULL;
--
-- Add field display_currency to examplemodelwithlateraddedmoneyfield
--
ALTER TABLE "core_examplemodelwithlateraddedmoneyfield" ADD COLUMN "display_currency" varchar(3) DEFAULT 'XYZ' NOT NULL;
ALTER TABLE "core_examplemodelwithlateraddedmoneyfield" ALTER COLUMN "display_currency" DROP DEFAULT;
COMMIT;
```
The problematic part is `DEFAULT 'XYZ' NOT NULL` — I assume it should be just `NULL`? Either way, in the database the field ends up being non-nullable.

From what I've seen it should behave just as a regular CharField, so it could be related to Django version. I've tested on 2.2.7, in an example repo I'm preparing, and a few other 2.2.* versions when I first encountered it.

So far reproduced with:
Django: 2.2.*
Django Money: 0.15.1 -- 1.0
Python: 3.7.3 -- 3.8.0

Got any directions for me?

jrocketfingers:
When setting `default_currency=None` django migrations would
erroneously make the `CurrencyField` non-nullable. This is because
Django's `Field.clone()` uses `deconstruct()` to clone the field
during migrations. Since `None` is actually user set (and can differ
from the default), we need to explicitly add it to kwargs.

Fixes #530.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -294,6 +294,8 @@ class MoneyField(models.DecimalField):
             del kwargs["default"]
         else:
             kwargs["default"] = self.default.amount
+        if self.default_currency is None:  # None needs to be preserved
+            kwargs["default_currency"] = self.default_currency
         if self.default_currency is not None and self.default_currency != DEFAULT_CURRENCY:
             kwargs["default_currency"] = str(self.default_currency)
         if self.currency_choices != CURRENCY_CHOICES:
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,15 @@
 Changelog
 =========
 
+`1.1.dev`_ - (unreleased)
+-------------------------
+
+Fixed
+~~~~~
+
+- Fix ``default_currency=None`` not making ``CurrencyField`` nullable in SQL migrations `#530`_ (`jrocketfingers`_)
+
+
 `1.0`_ - 2019-11-08
 -------------------
 


jrocketfingers:
The migration tests do not cover SQL migrations from what I can tell (nor does it make sense, given that we'd have to cover several backends). Any suggestions on how to prove this works?

djmoney/models/fields.py
```diff
@@ -294,6 +294,8 @@ def deconstruct(self):
             del kwargs["default"]
         else:
             kwargs["default"] = self.default.amount
+        if self.default_currency is None:  # None needs to be preserved
+            kwargs["default_currency"] = self.default_currency
```
jrocketfingers: I could, alternately, do:
```
kwargs["default_currency"] = self.default_currency
...
if self.default_currency is not None and self.default_currency != DEFAULT_CURRENCY:
    kwargs["default_currency"] = str(self.default_currency)
```

benjaoming:
> The migration tests do not cover SQL migrations from what I can tell (nor does it make sense, given that we'd have to cover several backends). Any suggestions on how to prove this works?

This is great! Thanks for your thorough investigation! To me, it should be enough with the code comment to avoid regressions. As such, no need for a regression test.

I don't know if a regression test could test that a nullable field is indeed nullable? But as I said, I don't find it necessary.

If you could add a changelog entry, that'd be great. Would need a new headline `1.1.dev (unreleased)` entry to be added (denoting in the changelog what is being worked out in the current prerelease series).

jrocketfingers:
No problem, I'm just glad this is god damn fixed -- that took some elbow grease to figure out 😵

Added the changelog entry, please check the format. I wasn't sure should I mention the Issue, or the PR. Also, let me know do you prefer the current version, or https://github.com/django-money/django-money/pull/532#discussion_r351478931.

djmoney/models/fields.py
benjaoming: Am not sure I understand the benefits of this? Why is `str()` necessary here and not in the alternative solution?

jrocketfingers: `str()` is the original code, and my guess is that it's there to cast Currency objects. If we were to coerce `None`, it'd say `"None"` in the DB, as if it was currency.

I realize now that it's not clear from looking at the line I commented on, but take a look at the whole segment to get a feel for it.

benjaoming: If it's preferable to coerce to a string, then that should be the solution? I understand that it shouldn't be done for `None` :)

jrocketfingers: Just to clarify, that == the current version?

benjaoming: Ah no, I was sensing that you preferred your own alternative suggestion and not the current PR's solution? Could you maybe clarify which one of them you prefer?

jrocketfingers: None, hence the question :) Let's keep the one that's currently used then.

But, **do not merge this yet**. It might be the cause of inability to migrate to a nullable currency field. I _think_ it's tied to my specific usecase (I've already migrated non-nullable currency and am trying to use this fix to make it nullable), but I can't be sure.

It essentially changes the migration state mid-migration and makes the migration executor think that the currency field was already nullable. It's absolutely cursed.

I'm preparing an investigation report I'll be posting as soon as I'm a bit clearer on the cause.

benjaoming: Maybe it's better to make a test case then?

jrocketfingers: From what I'm seeing, yes. No clue how to best do it, though, since it _might_ need to be at the SQL level, _or_ assert that the resulting db fields are actually nullable.

benjaoming: > assert that the resulting db fields are actually nullable.

I think that sounds like a good test-case that doesn't need to introspect the database.

benjaoming: You could add a new field/model here: https://github.com/django-money/django-money/blob/master/tests/testapp/models.py

jrocketfingers:
The current solution works, but doesn't get us all the way.

What happens when you want to migrate to a nullable field:
- Migration Operation 1: `AlterField` for `MoneyField` should set `default_currency=None`
  - no effect on the database, but migration state is affected
  - `MoneyField.contribute_to_class` is called in the process (while rendering state models, IIRC), which adds a `CurrencyField` with `null=True` to the state
  - By the time `CurrencyField.contribute_to_class` gets called directly `CurrencyField` already exists -- so it gets skipped and doesn't set `null=False` correctly
- Migration Operation 2: `AlterField` for `CurrencyField` should set `null=True`
  - everything goes as if it was a regular `CharField`, _but_, `from_state` now has the state after Operation 1
  - that means the `from_state` has `null=True` on the model, the same as `to_state`
  - this is interpreted as a no-op, nothing happens

An astute observer might notice that we don't really need Operation 1. What happens if we remove it?
- Migration Operation 1a: `AlterField` for `CurrencyField` should set `null=True`
  - `MoneyField.contribute_to_class` is called first, and again, adds `CurrencyField` that has `null=False`, and `default=<settings value>`, since the previous (starting) state has a `MoneyField` with no `default_currency=None`
  - `CurrencyField.contribute_to_class` is skipped again, for the same reason as above
  - no operation is performed as there were no changes to be seen by migration executor

My brain is a mush at this point, I'll re-read this and try to edit for clarity.

I don't have a solution, but a vague idea of what _could_ be done: `contribute_to_class` on `MoneyField` should be made aware it's in a migration and shouldn't add the currency field into the mix, allowing the CurrencyField to represent itself. This could cause other havoc I'm not aware of at the moment.

Part of me thinks forcing people to explicitly declare a currency field, yet have the money field magically include both would've been a cleaner solution for internal plumbing.

djmoney/models/fields.py
jrocketfingers: Will do, though we should probably move out of this review thread. I'll resolve it now, let's continue on the PR  itself.

benjaoming:
Your patience and systematic approach is admirable. I hope you get to the bottom of this.

> Part of me thinks forcing people to explicitly declare a currency field, yet have the money field magically include both would've been a cleaner solution for internal plumbing.

All doors may be open. If it's reasonable, we could deprecate `default_currency` and replace it with for instance `currency_field="field_name"`. That way, multiple fields could share the same currency (just saying).

jrocketfingers:
Thanks, though I'm at my wits' end :/ Won't have enough time to see this through to the end in the next two weeks, but I'll try to squeeze the tests at least.

> If it's reasonable, we could deprecate default_currency and replace it with for instance currency_field="field_name". That way, multiple fields could share the same currency (just saying).

I _really_ like this idea. Having more explicit relationships should simplify things internally. At the very least we'll avoid relying on `contribute_to_class`, which is from what I understand a private interface and might not be highest priority for Django team to maintain backwards compatibility on. Especially since I can't find the official doc describing the intended use of the method :)

There seems to be a need for it too: https://github.com/django-money/django-money/issues/488, https://github.com/django-money/django-money/pull/492

benjaoming:
Ah yes, I had forgotten about #488 - great find. I'm sure this work can be combined with the need for an alternative to (or replacement of!) `default_currency`. Would you be able to use that PR to speed up the process? As I see it, @Stranger6667 has already reviewed this work. What is needed would perhaps be:

1. Test consolidation and Python 2.7 support
1. A deprecation warning for `default_currency` since the better alternative is to define the `CurrencyField` and set the default there.

flaeppe:
I missed this PR completely when setting up #647 . Anyways, #647 includes a test for the deconstruction.

benjaoming:
Closing this in favour of #647, sorry that this one got stuck.

@jrocketfingers your input on #647 is highly appreciated in case something is missing!

jrocketfingers:
No worries! I should've pushed for it ¯\_(ツ)_/¯ I've taken a look, it looks good, no comments from my side 👍

PR #535, opened 2019-12-17
Import FieldDoesNotExist exception from current location

jeremy-engel:
The backwards-compatibility import from django.db.models.fields will be
removed in django 3.1.
See https://github.com/django/django/commit/129583a0d3cf69b08d058cd751d777588801b7ad#diff-bf776a3b8e5dbfac2432015825ef8afeL15-L18

### DIFF
--- a/djmoney/models/managers.py
+++ b/djmoney/models/managers.py
@@ -1,9 +1,9 @@
 from functools import wraps
 
+from django.core.exceptions import FieldDoesNotExist
 from django.db.models import Case, F, Q
 from django.db.models.constants import LOOKUP_SEP
 from django.db.models.expressions import BaseExpression
-from django.db.models.fields import FieldDoesNotExist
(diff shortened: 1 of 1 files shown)


benjaoming:
Thanks @jeremy-engel - do you have insights about django-money's Django 3.0 compatibility?

benjaoming:
Supposing that the `django_master` builds now target 3.1

jeremy-engel:
This should be the only change required for 3.1. I just double-checked the removals in https://docs.djangoproject.com/en/dev/releases/3.1/ and did not see anything else. Tests pass in this repo for current django master with python 3.7, and I haven't run into any issues with projects that use django-money, so it should be ready for that release when it comes.

benjaoming:
That's so great to know - thanks for checking this @jeremy-engel - I guess we can declare Django 3.1 support in a patch release even :tada:

PR #536, opened 2019-12-17
Add Django 3.0 to Travis

benjaoming:
Maybe Django 3.0 corresponded to `master` when the Travis matrix was changed last time? But I doubt it.


benjaoming:
This seems to work well!

PR #541, opened 2020-02-14
Optimize money operations on moneys with the same currencies

horpto:
if AUTO_CONVERT_MONEY is enabled many operations will be slower. It's ok for values with different currencies. But for the same currencies it's quite strange and unjustifiably (check on performance comparison).

```
from time import perf_counter

m = Money('0', 'RUB')
s = Money('1', 'RUB')

t0 = perf_counter()
for i in range(1000):
    m += s

print(perf_counter() - t0)  # 0.5836477298289537

m = Money('0', 'RUB')

t0 = perf_counter()
for i in range(1000):
    m.amount += s.amount

print(perf_counter() - t0)  # 0.0011850465089082718
```

### DIFF
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -90,7 +90,7 @@ def maybe_convert(value, currency):
     """
     Converts other Money instances to the local currency if `AUTO_CONVERT_MONEY` is set to True.
     """
-    if getattr(settings, "AUTO_CONVERT_MONEY", False):
+    if getattr(settings, "AUTO_CONVERT_MONEY", False) and value.currency != currency:
         from .contrib.exchange.models import convert_money
 
         return convert_money(value, currency)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,6 +5,11 @@ Changelog
 
 `1.1`_ - TBA
 
+Fixed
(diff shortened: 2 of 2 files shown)


benjaoming:
@horpto thanks! could you add a change log entry for this, same format as current change log entries?

https://github.com/django-money/django-money/blob/master/docs/changes.rst

horpto:
@benjaoming done

Stranger6667:
Hello @horpto !

Thank you for bringing this up, a very precise point :)
I do agree with the optimization, absolutely. But the code you provided in the second part of your comparison doesn't invoke `maybe_convert` and operates with decimal values directly. 
A more fair comparison would have the first part of the code executed with and without your patch. On my machine it is:

- Master branch: 0.025685272999908193
- Patched: 0.0026495460006117355

Which is still an impressive ~10x speedup! :)

benjaoming:
@Stranger6667 I guess that performance changes can be measured and amplified by the level of which the call is made?

Stranger6667:
@benjaoming yep, I believe the most visible changes will be for cases when some aggregation  (with `sum` or manually with `+`) is done on a big amount of instance

PR #554, opened 2020-06-06
fix: inherit decimal places from arguments in arithmetic operations #522

### LINKED ISSUE #522: Calculation with money fields leads to inconsistent `decimal_places`
Related to #521 I realized that the decimal-places are ignored when calculations are done:

```python

In [45]: one = Money(2.000000005, "USD", decimal_places=10)

In [46]: two = Money(1.0000, "USD", decimal_places=4)

In [47]: three=one+two

In [48]: three
Out[48]: <Money: 3.000000005 USD>

In [49]: three.decimal_places
Out[49]: 2
```

As far as I can see the decimal places are always set back to the `DEFAULT_DECIMAL_PLACES` from the settings. This seems inconsistent.

wearebasti:
This is supposed to resolve #522 

The `DefaultMoney` creates new `DefaultMoney`-Objects in the operations and returns those. These are initialized with the default decimal_places instead of the ones from the given operators.

To make sure to not "cut" decimal_places the 'biggest' one is taken into account.

I am not a big fan of the code duplication I introduced but I didn't see a nicer way (for now).

Added the explicit linking of `__ruml__` to `__mul__` as they are set like this in the `DefaultMoney` class but not in `Money` leading to inconsistent behaviour (a*b != b*a)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,15 @@
 Changelog
 =========
 
+`1.2`_ - TBA
+------------
+
+Fixed
+~~~~~
+
+- Resulting Money object from arithmetics (add / sub / ...) inherits maximum decimal_places from arguments `522` (`wearebasti`)
+
+
 `1.1`_ - 2020-04-06
 -------------------
 
--- a/tests/test_money.py
+++ b/tests/test_money.py
@@ -42,3 +42,45 @@ def test_round():
 def test_configurable_decimal_number():
     # Override default configuration per instance
     assert str(Money("10.543", "USD", decimal_places=3)) == "$10.543"
+
+
+def test_add_decimal_places():
+    one = Money(1.0000, "USD", decimal_places=4)
+    two = Money(2.000000005, "USD", decimal_places=10)
+
+    result = one + two
+    assert result.decimal_places == 10
+
+
+def test_add_decimal_places_zero():
+    two = Money(2.005, "USD", decimal_places=3)
+
+    result = two + 0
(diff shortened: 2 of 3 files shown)


tests/test_money.py
```diff
@@ -42,3 +42,45 @@ def test_round():
 def test_configurable_decimal_number():
     # Override default configuration per instance
     assert str(Money("10.543", "USD", decimal_places=3)) == "$10.543"
+
+
+def test_add_decimal_places():
+    one = Money(1.0000, "USD", decimal_places=4)
+    two = Money(2.000000005, "USD", decimal_places=10)
```
Stranger6667, line 49: Generally, I'd avoid creating `Money` instances from `float` across the codebase I suggest using strings instead

wearebasti: You're absolutely right there, changed it to strings. Was a bit lenient in the tests as DefaultMoney puts it in a string in the constructor...

docs/changes.rst
```diff
@@ -3,6 +3,15 @@
 Changelog
 =========
 
+`1.2`_ - TBA
+------------
+
+Fixed
+~~~~~
+
+- Resulting Money object from arithmetics (add / sub / ...) inherits maximum decimal_places from arguments `522` (`wearebasti`)
```
benjaoming, line 12:
```suggestion
- Resulting Money object from arithmetics (add / sub / ...) inherits maximum decimal_places from arguments `#522`_ (`wearebasti`_)
```

Need to add the link URLs, too.

Issue #521, opened 2019-10-24
Broken localization with 0.15.1


wearebasti:
With the merge of #501 the Money objects are automatically using the `decimal_places` of the underlying DB-`MoneyField` (if it exists).

This leads to weird localization behaviour in templates. We use a rather high number of decimal places to ensure correct calculations while displaying rounded localized values to the users.

Expected behaviour would be imo that by default the localization should use 2 decimal places unless told otherwise.

Can we achieve a "short" localization somehow or is this a bug?

Very short example:
```python
In [42]: y=Money(2.0000000005, "EUR", decimal_places=10)

In [43]: y
Out[43]: <Money: 2.0000000005 EUR>

In [44]: str(y)
Out[44]: '2,0000000005 €'
```

benjaoming:
> Very short example:

To understand you correctly: Is this an example of the ideal/expected behavior or the bug? :)

Your example performs as my "expected behavior"...

> Expected behaviour would be imo that by default the localization should use 2 decimal places unless told otherwise.

In the example, `decimal_places` is set to `10` for this particular `Money` instance and the human-readable representation takes this into account. Over-ruling this with a default setting would be strange, when this particular `Money` instance explicitly deviates from this.

wearebasti:
That was just for 'showing' that it's taken into account in the localization/print. (In a nutshell: the `Money(..., decimal_places=10)` should be the instantiation from a `MoneyField`)

The real problem: If the `Money` instance is created from a `MoneyField` it is taking the decimal-places attribute from the `MoneyField` into account.

Which leads to 'realy high' numbers of decimal-places in the `Money` instance while we only want to use it for calculations & persistence.

Does this clarify things for you?

benjaoming:
I understand that it is desirable to be able to calculate with more decimals than what is displayed, yes.

What I'm saying is that to me, the way that the field is represented as a string, can be argued to be just as correct when using the instantiated `decimal_places` value instead of the django project default.

Would it be better to have a way to configure default + localized display independent of `decimal_places`? Say if we had `DECIMAL_PLACES` and a new `DECIMAL_PLACES_DISPLAY` as a configurable list of currencies and their desired decimal places?

    DECIMAL_PLACES_DISPLAY = [('USD', 2)]

benjaoming:
Ah okay, let me re-iterate the draft of a new setting:

```python
DECIMAL_PLACES = 2
DECIMAL_PLACES_DISPLAY = [
[...]

PR #555, opened 2020-06-07
Fix: provide option to set decimal_places_display explicitly #521

wearebasti:
Based on the discussion with @benjaoming I implemented the settings dictionary mapping currency to decimal-places-displayed.

It's been a while since the issue was created, hopefully I still had the 'right' solution in my head. What do you think?


docs/changes.rst
```diff
@@ -3,6 +3,14 @@
 Changelog
 =========
 
+`Unreleased`_
+-------------
+
+Added
+~~~~~
+- Possibility to set localized string output to length of `decimal_places_display` `#521`_ (`wearebasti`)
```
benjaoming, line 11:
```suggestion
- Improved localization: New setting ``CURRENCY_DECIMAL_PLACES_DISPLAY`` configures decimal places to display for each configured currency `#521`_ (`wearebasti`)
```

benjaoming:
@Stranger6667 would you want to review?

Stranger6667:
Looks good to me!

wearebasti:
@benjaoming I merged master into this branch; build works again.

In regards to the semantics: I'd go with the "default" as you proposed and added a setter for the `decimal_places_display` property, this way the user can set the property "after the fact".

```python
# given `DECIMAL_PLACES_DISPLAY` is 2 for USD
a = Money("2", "USD")
b = Money("1.234567", "USD", decimal_places_display=5) 

c = a + b
print(str(c))
> $3.23

c.decimal_places_display = 5
print(str(c))
> $3.23457
```

I guess that's a good start for now and if somebody has different requirements that would be something to evolve with new features?

djmoney/money.py
```diff
@@ -24,8 +24,21 @@ class Money(DefaultMoney):
 
     def __init__(self, *args, **kwargs):
         self.decimal_places = kwargs.pop("decimal_places", DECIMAL_PLACES)
+        self._decimal_places_display = kwargs.pop("decimal_places_display", None)
         super().__init__(*args, **kwargs)
 
+    @property
+    def decimal_places_display(self):
+        if self._decimal_places_display is None:
+            self._decimal_places_display = DECIMAL_PLACES_DISPLAY.get(self.currency.code, self.decimal_places)
```
Stranger6667, line 33: I am not sure about caching on the attribute level. What if somebody will use `override_settings(CURRENCY_DECIMAL_PLACES_DISPLAY=...)` in their test, but a test money object is set up and this property is evaluated before the decorator is applied? In this case, it might be confusing - no explicit `decimal_places_display` in the object, but the settings change won't affect it. It is a kinda special example but might cause problems for users

djmoney/money.py
```diff
@@ -24,8 +24,21 @@ class Money(DefaultMoney):
 
     def __init__(self, *args, **kwargs):
         self.decimal_places = kwargs.pop("decimal_places", DECIMAL_PLACES)
+        self._decimal_places_display = kwargs.pop("decimal_places_display", None)
```
Stranger6667, line 27: Probably having `decimal_places_display` in the signature will make this argument more explicit (also, IDEs might pick it up). E.g.

```suggestion
    def __init__(self, *args, decimal_places_display=None, **kwargs):
        self.decimal_places = kwargs.pop("decimal_places", DECIMAL_PLACES)
        self._decimal_places_display = decimal_places_display
```

I believe it should work on the supported Python versions

wearebasti: good idea - changed it

djmoney/money.py
wearebasti: I usually wouldn't change production code for the sake of test-code, but I also do not see a specific reason to keep the local caching - so I removed it :)

PS: I was actually wondering if the override settings would work without the cache as the `from settings import DECIMAL...` thing probably keeps its value?

Stranger6667: You are right - I overlooked it :(

wearebasti:
I've just seen that TravisCI is on "Pending" - is there something I can (or have to) do to get it "green"?

docs/changes.rst
```diff
@@ -3,8 +3,13 @@
 Changelog
 =========
 
-`1.2`_ - TBA
-------------
+`Unreleased`_
```
benjaoming, line 6:
```suggestion
`1.2`_ - TBA
```
This needs to be revered, 1.2 is already a known link, I think the other one will break the docs build.

docs/changes.rst
```diff
@@ -3,8 +3,13 @@
...
+-------------
```
benjaoming, line 7:
```suggestion
------------
```

benjaoming (approved):
Changing docs stuff, otherwise supposing everything is still fine

benjaoming:
@wearebasti sorry that this is taking so long!

wearebasti:
> @wearebasti sorry that this is taking so long!

@benjaoming all good, thanks for all your work!

Would appreciate a release (including this fix) so I don't have to pin versions anymore :)

benjaoming:
Thanks so much for the contribution @wearebasti - and deeply sorry about the waiting time! As you can see, we are a bit low on maintenance resources.

PR #557, opened 2020-06-17
set default to NOT_PROVIDED

tned73:
Thanks for the great work on Django-Money.

Our usage of the MoneyField in our models requires that it can be mixer/faked for test purposes. This is currently not possible, because the default is None and mixer will assume that it was set intentionally and would not change/fake it.
With the correct change that the field no longer defaults to 0, a lot of our test failed. This was caused by the fact that the default is now None.
The mixer package presumes that if the default is NOT_PROVIDED, then there was no default value in the definition of the field on the model. In that case it will generate an proper value for the field when the model instance was created by mixer.blend(). This allows us to let mixer fill in the non trivial required MoneyFields.
an extra issue arose in the migration tests when the default was now NOT_PROVIDED, this triggered the makemigration to check if null was allowed and if it was not an initialisation migration, it would correctly ask for a one-time value so it could update instances if need be.

I hope that my contribution is appreciated, if any questions or remarks arise, I will be happy to answer them.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -274,12 +274,18 @@ class MoneyField(models.DecimalField):
         else:
             return super().get_default()
 
+    @property
+    def _has_default(self):
+        # Whether the field has an explicitly provided non-empty default.
+        # `None` was used by django-money before, and we need to check it because it can come from historical migrations
+        return self.default not in (None, NOT_PROVIDED)
+
     def formfield(self, **kwargs):
         defaults = {"form_class": forms.MoneyField, "decimal_places": self.decimal_places}
         defaults.update(kwargs)
         defaults["currency_choices"] = self.currency_choices
         defaults["default_currency"] = self.default_currency
-        if self.default is not None and self.default is not NOT_PROVIDED:
+        if self._has_default:
             defaults["default_amount"] = self.default.amount
         return super().formfield(**defaults)
 
@@ -290,9 +296,7 @@ class MoneyField(models.DecimalField):
     def deconstruct(self):
         name, path, args, kwargs = super().deconstruct()
 
-        if self.default in [None, NOT_PROVIDED]:
-            pass
-        else:
+        if self._has_default:
             kwargs["default"] = self.default.amount
         if self.default_currency is not None and self.default_currency != DEFAULT_CURRENCY:
             kwargs["default_currency"] = str(self.default_currency)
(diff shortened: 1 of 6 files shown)


benjaoming:
Thanks for this! I hadn't seen the PR before. I don't mind assisting to have it merged, but I don't have so much bandwidth and I think maybe it's the same for @Stranger6667 

I would like to ask this:

How does changing this default affect current deployments of django-money? Do they need new migrations?

tned73:
Hi @benjaoming,

I missed your response. I will check the update strategy as you requested and come back with the results.

tned73:
@benjaoming 
I found a small bug in this PR and updated tests.

With regards to makemigrations, if there are migrations which has no `default=None` on the Field defined, but has default=None in the last migration it will make a new migration. That is the only case I found where this fix triggers a new migration.

With a model class with:
`money = MoneyField(max_digits=7, null=True)`
could have a historic migration like this (but not always having the default=None part), due to the implicit default=None of the field definition
`field=djmoney.models.fields.MoneyField(decimal_places=2, default=None, max_digits=7, null=True)),`
after this fix the makemigration will create a migration without de default
`field=djmoney.models.fields.MoneyField(decimal_places=2, max_digits=7, null=True),`
.
As for current deployments it does not change the operation. The same logic will still be executed. if a money field is defined with `null=False` or `null=True` and you create a instance without defining the money value, it will still try to insert None in the DB. Depending on the allowance of writing None in the record you could get an Integrity exception. This is no difference as the current operation. 
The only benefit of having the default set to not_provided is that mixer (and maybe other test systems) don't assume you have defined the field with an explicit `default=None` and therefore try to make up a test value.

An side effect of this change is that it also solves a bug: if you change a money field from null=True to null=False and you have not defined a default, with this PR makemigrations will ask for a default value for the historic records where as the current implementation presumes you defined default=None on the Field.

Stranger6667 (approved):
@tned73 Thank you for your contribution! Looks good to me! This change makes the `MoneyField` more in line with the default behavior of other Django fields (they have `default=NOT_PROVIDED`

I will add minor changes in subsequent commits (reduce branching & changelog entry).

As for `mixer` - in the future, I'd like to use the Hypothesis' [Django plugin](https://hypothesis.readthedocs.io/en/latest/django.html) for data generation, but I'll compare them first (they may be not mutually exclusive).

Stranger6667:
Released in `1.3`

PR #604, opened 2021-02-02
Do not alter existing objects.

### LINKED ISSUE #603: djmoney alters existing objects
We have an application that uses `moneyed` and (sub-classes of) its `Money` type in a non-Django environment. We recently started integrating this into a Django app that uses djmoney and started getting weird errors in seemingly unrelated cases  because instances of `djmoney.money.Money` started showing up there.

We tracked this down to a behavior of the djmoney package in `djmoney.models.fields`:
https://github.com/django-money/django-money/blob/bb222ed619f72a0112aef8c0fd9a4823b1e6e388/djmoney/models/fields.py#L38
https://github.com/django-money/django-money/blob/bb222ed619f72a0112aef8c0fd9a4823b1e6e388/djmoney/models/fields.py#L209

The two places alter **the type of an existing object**. Can we change this to copying the objects before altering them?

moser:
My proposal to fix #603

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -35,7 +35,7 @@ def get_value(obj, expr):
     else:
         expr = expr.value
     if isinstance(expr, OldMoney):
-        expr.__class__ = Money
+        expr = Money(expr.amount, expr.currency)
     return expr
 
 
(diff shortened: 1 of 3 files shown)


benjaoming:
Hi @moser - you should be able to get the tests passing by merging in the latest master branch :+1:

benjaoming (approved):
Thanks for this contribution! LGTM - the code that's fixed looks like it's a leftover from some other intentions, seems wrong to alter an existing object like the old code did.

Stranger6667:
I absolutely agree with these changes - it was a wrong decision to mutate the input. It will be nice to have a regression test + a changelog entry

Stranger6667:
I will take care of these :)

PR #605, opened 2021-02-03
Remove reversion, it does not work with Django 4 (alpha)

benjaoming:
Trying to fix test error in #604


benjaoming:
Since this fixes a current issues with tests, am a bit keen on merging it. @Stranger6667 can you tell me if anything should be done after this? :pray:

Stranger6667:
Looks good to me!

PR #610, opened 2021-04-11
Fix serializer with model field

melvyn-oviavo:
When using the currency_field_name option to MoneyField, pass the field instance to the resolver. If we don't do this, we can't load the fixtures as it will be looking for 'field_name_currency' in the fixture.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,8 +81,6 @@ jobs:
           - python: 3.6
             tox_env: django31-py36
           - python: 3.6
-            tox_env: django_master-py36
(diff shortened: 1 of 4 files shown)


benjaoming (approved):
Thanks for fixing the "footshooting", @melvyn-oviavo. I hope the CI wasn't too frustrating of an experience :)

The changes look great! Thanks for contributing!

melvyn-oviavo:
>  I hope the CI wasn't too frustrating of an experience

Neh, no worries. Thanks for the quick merge. 👍

Issue #567, opened 2020-11-09
Support most recent changes in py-moneyed for formatting


spookylukey:
In [py-moneyed](https://github.com/limist/py-moneyed) we have just merged a change that deprecates `moneyed.localization` in favour of `moneyed.l10n`. The latter contains a single, new `format_money` function, which simply delegates to [Babel number formatting functions](https://babel.pocoo.org/en/latest/api/numbers.html). This fixes many long-standing issues in our formatting, and frees us of the need to try to manually define all the different formats - we get them from Babel, which is based on [CLDR](http://cldr.unicode.org/) data.

Before releasing a new version of `py-moneyed`, I'd like to check that the new API works for django-money, since I know it is a major user of py-moneyed. Would it be possible for you to do an experimental branch that doesn't use `moneyed.localization` at all (or at least, only as a fallback)?

Please see the [new docs](https://py-moneyed.readthedocs.io/en/latest/formatting.html) and [CHANGES.rst](https://github.com/limist/py-moneyed/blob/master/CHANGES.rst)

Thanks.

Stranger6667:
I was hoping for this for years :) Finally, it is happening. I think that it is a great change and I'd be happy to have it in `django-money`. What do you think, @benjaoming ?

spookylukey:
FYI - I have released py-moneyed 1.0 now, without waiting for a reply here, as people have been waiting a few months. I don't think you should have too much problem integrating the changes.

antonagestam:
Hi! I started looking into this as it seems like the best issue to work on to try to get django-money on par with py-moneyed. I almost immediately bumped into an issue that I haven't managed to figure out on my own.

In the [current implementation](https://github.com/django-money/django-money/blob/78e7aed33ba2601eacf320a2f766fec4e3e14249/djmoney/money.py#L100) of django-money's `Money.__str__()`, the call to `format_money()` includes a `decimal_places` kwarg that is set to `Money.decimal_places_display` which can either be read from settings or set directly on a `Money` instance.

However, in [the new](https://github.com/py-moneyed/py-moneyed/blob/master/moneyed/l10n.py#L10) babel-backed `format_money()` implementation there is no `decimal_places` argument. Babel does allow overriding the number of decimal places displayed, but it seems like it is only possible by overriding the format (and setting `currency_digits=False`). So it's a little unclear to me if the current behavior of allowing overriding the number of decimal places (without providing a formatting string) can be easily supported.

[...]

PR #612, opened 2021-04-19
Drop support for legacy versions

antonagestam:
- Drop support for Django 1.11, 2.1 and 3.0.
- Drop support for Python 3.5.
- Add support for Django 3.2.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
 
       - run: pip install tox
 
@@ -40,53 +40,27 @@ jobs:
       matrix:
         include:
           - python: pypy3
-            tox_env: django111-pypy3
-          - python: pypy3
-            tox_env: django21-pypy3
-          - python: pypy3
             tox_env: django22-pypy3
           - python: pypy3
-            tox_env: django30-pypy3
-          - python: pypy3
             tox_env: django31-pypy3
-          - python: 3.5
-            tox_env: django111-py35
-          - python: 3.5
-            tox_env: django21-py35
-          - python: 3.5
-            tox_env: django22-py35
-          - python: 3.6
-            tox_env: django111-py36
-          - python: 3.6
-            tox_env: django21-py36
           - python: 3.6
             tox_env: django22-py36
-          - python: 3.6
-            tox_env: django30-py36
           - python: 3.6
             tox_env: django31-py36
           - python: 3.6
             tox_env: no_rest_framework
           - python: 3.7
-            tox_env: django21-py37
-          - python: 3.7
             tox_env: django22-py37
-          - python: 3.7
-            tox_env: django30-py37
           - python: 3.7
             tox_env: django31-py37
           - python: 3.8
             tox_env: django22-py38
-          - python: 3.8
-            tox_env: django30-py38
           - python: 3.8
             tox_env: django31-py38
           - python: 3.8
             tox_env: django_master-py38
           - python: 3.9
             tox_env: django22-py39
-          - python: 3.9
-            tox_env: django30-py39
           - python: 3.9
             tox_env: django31-py39
           - python: 3.9
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
     hooks:
     -   id: pyupgrade
         args:
-        - --py3-plus
+        - --py36-plus
 
 -   repo: https://github.com/asottile/seed-isort-config
     rev: v2.2.0
--- a/README.rst
+++ b/README.rst
@@ -20,8 +20,8 @@ django-money
 A little Django app that uses `py-moneyed <https://github.com/py-moneyed/py-moneyed>`__ to add support for Money
 fields in your models and forms.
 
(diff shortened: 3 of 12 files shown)


Stranger6667 (changes requested):
There is also version info in `README.rst` and [setup.py](https://github.com/django-money/django-money/blob/master/setup.py#L84). Could you, please, update it as well?

antonagestam:
@Stranger6667 Oops, I missed those. I pushed a fixing commit 🙂

Stranger6667 (approved):
Great! I am thinking whether it makes sense to add support for `pymoneyed==1.x` before dropping support for unsupported Python / Django versions. This way, users on these versions may upgrade and use `babel`-powered formatting through `pymoneyed` without upgrading the rest from e.g. Django 1.11. What do you think?

antonagestam:
@Stranger6667 I would have definitely gone the other way and not go out of my way to build stuff for unsupported versions, I tend to think about it as cleaning up the kitchen before starting making dinner ;) **But**, I'm not the maintainer here and I'll go along in the direction you set. In this particular case, it might not be much extra work either way.

Btw, I bumped into an issue with upgrading the formatting, I'll post the details on that issue.

benjaoming:
@Stranger6667 sounds like a reasonable little intermediate release. Could that be added and released through a separate PR?

antonagestam:
@benjaoming For reference, there's been discussion about that [here](https://github.com/django-money/django-money/issues/567) 🙂 I'm thinking this PR can stay open until a new version is released with support for py-moneyed 1.x.

antonagestam:
@Stranger6667 @benjaoming Hi 👋 I've started working on the final step in the plan laid out, and [upgrading to py-moneyed 2.0](https://github.com/antonagestam/django-money/pull/1). Is there any chance to get this PR merged?

Stranger6667:
@antonagestam Hi! I'll take a look during the weekend :)

antonagestam:
Another thing for discussion here though, it would make sense to me also drop support for Python 3.6 now as it only has 3 more months to live, and probably even less by the time we're ready to release. It makes sense to me since we're about to make a major release anyway for py-moneyed 2. Wdyt?

Stranger6667:
Yep, looks good to me! I think if we declare support for Django 3.2, then it will be also nice to solve #615 in this release. Especially as the fix is a one-liner (as in [this PR](https://github.com/StevenMapes/django-money/pull/1)), so Django 3.2 users will have a warning-less experience :)

> Another thing for discussion here though, it would make sense to me also drop support for Python 3.6 now as it only has 3 more months to live, and probably even less by the time we're ready to release. It makes sense to me since we're about to make a major release anyway for py-moneyed 2. Wdyt?

I'd go for as frictionless experience as possible - I think we can just make another release when we drop Python 3.6 after its EOL.

Stranger6667:
On #615 - it surely might be a part of a different PR.

Stranger6667 (approved):
Feel free to squash & merge this PR once you feel it's ready :)

benjaoming:
In case we merged this now, I wonder if we have any good ways to ensure that we don't release from `master` before #615 is fixed :)

StevenMapes:
Yeah sure I just didn't finish the or when by browser went wrong then I
went on vacation

antonagestam:
@StevenMapes Great, I brought in your commit here 👍

@benjaoming @Stranger6667 Unless there are other concerns, I think this should be good to go now 🙂

Stranger6667:
I think we are good to go :)

benjaoming:
THANKS ALL :ship: 

I just went ahead and did the honors of merging, but we should be able to release this as well.

Suggesting that we bump to 2.1 because of the backwards incompats... any other suggestions?

antonagestam:
@benjaoming 🥳

On releasing, since these are breaking changes, doesn't that warrant a major version bump, e.g. 3.0? My plan was to prepare for further work with this PR, and follow that up with the py-moneyed upgrade (that is work in progress) in the same release. The reason I think it makes sense to have them both in 3.0 is that they are both breaking changes, and otherwise we'd have to release a 4.0 so soon after.

I'll try to find the time to finalize the [py-moneyed work](https://github.com/antonagestam/django-money/pull/1) soon so that it doesn't hold up the release! What do you think? 🙂

benjaoming:
@antonagestam it's still only smaller breaking changes, so since we are sticking to SemVer, it's still a minor version bump with no changes to API or data models and such.

I'll go ahead and get this released now :)

I think that it might work great for you to bump py-moneyed on the major digit, but once py-moneyed 2.0 is out, we can create a django-money release that bumps to py-moneyed >=2.0 and ALSO bump the major version of django-money to 3.0 since this could be considered a breaking change for those using django-money.

That leaves a nice upgrade path, too: Upgrade to django-money 2.1 and any Python/Django versions that might need upgrading, too, _then_ upgrade py-moneyed.

antonagestam:
@benjaoming Hmm, I'd definitely consider dropping support for Python versions as a breaking change though.

benjaoming:
@antonagestam it's definitely a breaking change, but the way that Django and Python are being released, most projects are choosing to treat this with minor version bumps.

I have gone ahead with the release now, so we are ready for the next py-moneyed :+1: :100: 

image

antonagestam:
@benjaoming Cool! Exciting to see things in motion! 😃

PR #614, opened 2021-04-24
Currency max length defaults

ayushin:
Attention - this affects initial migration, so CURRENCY_TICKER_MAX_LENGTH (default 3) can't be changed afterwards without creating a new migration

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,11 @@ Searching for models with money fields:
         BankAccount.objects.filter(balance__gt=Money(1, 'USD'))
         # Returns the "account" object
 
+The default currency code length is `3` but you can change it with the `CURRENCY_CODE_MAX_LENGTH` setting.
+
+Caution: this setting also affects the initial migration of the `exchange` plugin, so changing it after running
+the initial migration has no effect. (You'd need to `manage migrate exchange zero` and migrate again if you want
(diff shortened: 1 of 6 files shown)


ayushin:
This PR works for us

https://github.com/django-money/django-money/pull/614

benjaoming (changes requested):
Changes look good and shouldn't affect existing deployments. I don't understand why it's called "Currency Ticker"? It's not a term that I'm familiar with. Wouldn't we normally call this a "currency code"?

It would be nice to have a comment or some kind of documentation that this setting cannot be changed without manually creating database migrations.

ayushin:
I've changed it to CURRENCY_CODE I guess you are right, it is more suiting for the module naming.

Also added a note in the README

benjaoming (approved):
Thanks for addressing the feedback @ayushin - If you can add a release note for this, too, that would be sweet. Otherwise LGTM :+1:

Stranger6667 (approved):

benjaoming:
Hi @ayushin

You can fix the lint error with something like this:

```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

benjaoming (approved):
LGTM :rocket:

PR #619, opened 2021-05-20
Add support for py-moneyed 1.x

### LINKED ISSUE #567: Support most recent changes in py-moneyed for formatting

antonagestam:
Closes #567

### DIFF
--- a/djmoney/money.py
+++ b/djmoney/money.py
@@ -1,3 +1,7 @@
+import warnings
+from functools import partial
+from types import MappingProxyType
+
 from django.conf import settings
 from django.db.models import F
 from django.utils import translation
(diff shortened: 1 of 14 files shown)


Stranger6667 (approved):
Wow! LGTM! :) Thank you!

Stranger6667:
I'll add a changelog entry and will issue a new release :)

Stranger6667:
Oh, It seems like ``currency_digits`` isn't yet there - I will check the details tomorrow.

antonagestam:
@Stranger6667 You can set `currency_digits` and the other new options through `formatting_options`, so either of these:

```python
# in settings.py
MONEY_FORMAT = {
    "currency_digits": 3,
}

# in instantiation
Money(12.34, USD, format_options={"currency_digits": 3})
```

My intention with that is that we can add a `FormatOptions` TypedDict when adding type hints further on to get typing of the options. That type could also be exposed so that you could do:

```python
MONEY_FORMAT = FormatOptions(currency_digits=3)
```

antonagestam:
Should have added that to the description here, had a bit of lag between implementing and opening the PR ;)

PR #622, opened 2021-06-16
fix(validator): prevent attribute error

PedroBern:
Fix a bug where it was possible to get attribute error by calling `.amount` on a possible non-Decimal value.

### DIFF
--- a/djmoney/models/validators.py
+++ b/djmoney/models/validators.py
@@ -27,7 +27,10 @@ class BaseMoneyValidator(BaseValidator):
         if limit_value is None:
             return
         if isinstance(limit_value, (int, Decimal)):
-            cleaned = cleaned.amount
+            try:
+                cleaned = cleaned.amount
+            except AttributeError:
+                pass
         params = {"limit_value": limit_value, "show_value": cleaned, "value": value}
         if self.compare(cleaned, limit_value):
             raise ValidationError(self.message, code=self.code, params=params)


Stranger6667:
Hi! Thank you for your contribution! Could you, please, add a test case for it?

PedroBern:
Hi @Stranger6667  thanks for the quick response.

I was getting this error when using with DRF. My model has a minimum money validator of `Decimal(0)`, and when making a POST request with any value it was raising the error.

I fixed it by using this workaround, but not sure if I will be able to replicate it in your test cases.

I appreciate if it is merged because those new 3 lines will not impact negatively in anything.

Stranger6667:
Thank you @PedroBern for providing more context!

I'll check it in more details tomorrow / the day after and I think it could be released quite soon, but I want to have a regression case to avoid such faulty behavior in the future :)

PedroBern:
@Stranger6667  thank you very much :)

Stranger6667:
Unfortunately, I can't reproduce it locally with a config similar to what you mentioned. Could you, please share your model & serializer (at least partially)?

PedroBern:
@Stranger6667 thanks for investigating. Feel free to close then, I will be using my patch for now. Later I will try to make a reproducible example and share it here, if relevant we can reopen and merge.

benjaoming:
Closing - but please feel encouraged to re-open if you find the test case to reproduce.

PR #629, opened 2021-09-03
correct copy of attributes

tned73:
After an action like adding the attribute decimal_places_display was incorrectly set and it was interfering with formatting in the __str__. Even the deprecation warning was triggered.
This fix only copy the internal attributes and avoids incorrectly setting them.

Small proof of issue (Version 2.0.1) Also notice formatting problem.
```
> from djmoney.money import Money
> effe = Money("1.23", "EUR")
> print(effe)

€1.23

> print(effe._decimal_places_display)

None

> effe = effe + effe
> print(effe)

2.46 €

> print(effe._decimal_places_display)

2
```

### DIFF
--- a/tests/test_money.py
+++ b/tests/test_money.py
@@ -155,37 +155,45 @@ def test_sub_negative():
     assert total == Money(-33, "EUR")
 
 
-def test_add_result_in_proper_instance():
-    one = Money(1, "EUR")
-    assert one._decimal_places_display is None, "this should be None"
-    assert one.decimal_places == 2
-    two = Money(2, "EUR")
-    assert two._decimal_places_display is None, "this should be None"
-    assert two.decimal_places == 2
-    three = one + two
-    assert three._decimal_places_display is None, "this should be None as the others are None"
-    assert three.decimal_places == 2
+@pytest.mark.parametrize(
+    "decimal_places_display, decimal_places",
+    [
+        [None, None],
+        [0, 0],
+        [1, 0],
+        [4, 0],
+        [0, 1],
+        [1, 1],
+        [4, 1],
+        [0, 4],
+        [1, 4],
+        [4, 4],
+        [None, 4],
+        [None, 1],
+        [None, 0],
+        [4, None],
+        [1, None],
+        [0, None],
+    ],
+)
+def test_proper_copy_of_attributes(decimal_places_display, decimal_places):
+    one = Money(1, "EUR", decimal_places_display=decimal_places_display)
 
+    assert one._decimal_places_display is decimal_places_display
+    assert one.decimal_places == 2, "default value"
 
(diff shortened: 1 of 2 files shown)


tests/test_money.py
```diff
@@ -153,3 +153,39 @@ def test_sub_negative():
     for bill in bills:
         total -= bill
     assert total == Money(-33, "EUR")
+
+
+def test_add_result_in_proper_instance():
+    one = Money(1, "EUR")
+    assert one._decimal_places_display is None, "this should be None"
+    assert one.decimal_places == 2
```
Stranger6667, line 161: Generally, I'd prefer a bit higher level assertions (at least in new tests) - e.g. how values will be formatted, as these attributes are telling more about implementation details.

tests/test_money.py
```diff
@@ -153,3 +153,39 @@ def test_sub_negative():
...
+    two = Money(2, "EUR")
+    assert two._decimal_places_display is None, "this should be None"
+    assert two.decimal_places == 2
+    three = one + two
+    assert three._decimal_places_display is None, "this should be None as the others are None"
+    assert three.decimal_places == 2
+
+
+def test_add_copy_display_places():
```
Stranger6667, line 170: Also, it will be nice to use `pytest.mark.parametrize` as these 3 tests have similar structure

Stranger6667:
Thanks for working on this!

tned73:
you are welkom, I have changed the tests as best I can.  Let me know if it can be better.

tests/test_money.py
```diff
@@ -153,3 +153,47 @@ def test_sub_negative():
...
+    assert two.decimal_places == decimal_places
+
+    three = Money(3, "EUR")
+    one._copy_attributes(two, three)
+
+    assert three._decimal_places_display == decimal_places_display
+    assert three.decimal_places == max(2, decimal_places) if decimal_places is not None else 2
+
+    zero = Money(0, "EUR")
+    one._copy_attributes(Money(1, "EUR", decimal_places_display=3), zero)
```
Stranger6667, line 197: Oh, I thought you'd keep arithmetic operations and replace `assert three._decimal_places_display == ...` with assertions on string formatting (as in the PR description)

tned73: This is the function who has changed, by directly using this function in this test I had hoped to make the test more clear of what its testing purpose is. By using addition or subtraction it was masquerading it purpose.
The formatting randomness is a result of having the decimal_places_display incorrectly set and being used in selecting the format for str output.
So this fix is more about consistency then formatting, that is a nice extra result.

Stranger6667: Yep, makes sense

Stranger6667:
Released in 2.0.2

PR #638, opened 2021-10-02
Upgrade to py-moneyed 2.0 

antonagestam:
- Update py-moneyed to 2.0.
- Remove the deprecated `Money.decimal_places_display` property and argument.
- Remove the deprecated `CURRENCY_DECIMAL_PLACES_DISPLAY` setting.

_WIP_, there are some things to figure out in terms of behavior due py-moneyed no longer having a default dummy currency. I believe this should be pretty straightforward though.

Depends on https://github.com/django-money/django-money/pull/612


setup.py
```diff
@@ -66,7 +66,7 @@ def find_version():
     maintainer_email="greg@reinbach.com",
     license="BSD",
     packages=find_packages(include=["djmoney", "djmoney.*"]),
-    install_requires=["setuptools", "Django>=2.2", "py-moneyed>=1.2,<2.0"],
+    install_requires=["setuptools", "Django>=2.2", "py-moneyed>=2.0"],
```
pauloxnet, line 69: Maybe here we can be conservative to the next major `py-moneyed` version ?
```suggestion
    install_requires=["setuptools", "Django>=2.2", "py-moneyed>=2.0,<3.0"],
```

djmoney/money.py
```diff
@@ -175,22 +143,12 @@ def __rmod__(self, other):
...
-    if locale in moneyed.localization._FORMATTER.formatting_definitions:
-        return locale
-
-    return ""
+def get_current_locale():
+    return translation.to_locale(
+        translation.get_language()
+        # get_language can return None starting from Django 1.8
+        or settings.LANGUAGE_CODE
+    )
```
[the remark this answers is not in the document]
antonagestam: It probably can, but almost the entire function is botched here so it felt like a good chance to simplify it.

setup.py
antonagestam: In my experience that isn't common practice for package dependencies and might cause more trouble than its worth, i.e. it will put users in impossible situations if they want to upgrade py-moneyed before there is an explicitly compatible version of django-money.

Just two quick examples:

- https://github.com/samuelcolvin/pydantic/blob/master/setup.py#L136
- https://github.com/django/django/blob/main/setup.cfg#L40

flaeppe:
> there are some things to figure out in terms of behavior due py-moneyed no longer having a default dummy currency

@antonagestam Maybe I'm running ahead here, but I'm curious if you thought of a replacement value of `MoneyField(..., default_currency=DEFAULT_CURRENCY)`?

I'd guess a natural replacement could be `MoneyField(..., default_currency=None)`. Without any other adjustments that would lead to a nullable `CurrencyField` and without using option `MoneyField(..., null=True)` the "default" setup would have a non-nullable money-column. And looking at this https://github.com/django-money/django-money/issues/530#issuecomment-558081994 regarding integrity options the "invalid" case is hit as default(which is controlled by `MoneyField`'s descriptor class).

Perhaps `null=True` for `CurrencyField` should then instead be based on the `MoneyField(..., null=True)` option? This leads to the default setup of `MoneyField` with a non-nullable `CurrencyField` and non-nullable money-column as default. Thus requiring both a currency and amount to save a `MoneyField`, which feels like a natural default (inserting with the default currency would error on the not null constraint).

Again, looking at this https://github.com/django-money/django-money/issues/530#issuecomment-558081994 the exceptional case to support would be nullable money and non-nullable currency, which doesn't sound like a regular case (but maybe I'm wrong). But I'm thinking that maybe one could refrain from supporting that implicitly and instead support that via being explicit and do:

```python
class MyModel(models.Model):
    money_currency = CurrencyField(..., null=False)
    money = MoneyField(null=True, currency_field_name="money_currency", ...)
```

And then just ensure precedence on how the `CurrencyField` is set up is working correctly (think explicit currency field is already supported).

These changes could have _a lot_ of impact on a lot of projects though, migration changing wise. But perhaps there's a more graceful approach.

davidszotten:
hi. can i do anything to help move this forward?

antonagestam:
@davidszotten I'll get to it eventually, there's really nothing holding this back other than someone cranking through it.

davidszotten:
> @davidszotten I'll get to it eventually, there's really nothing holding this back other than someone cranking through it.

what are you referring to by "cranking through it"? (and is this something i could help with). e.g. is the main blocker  design decisions (like the one above about the default dummy currency) or fixing failing tests or something else?

antonagestam:
@davidszotten I'm saying there really aren't any blockers.

Yes, one has to take a stance on the question posed by @flaeppe, but I think that already comes naturally by a new default value for `DEFAULT_CURRENCY` which should propagate. (I don't have this entirety of the current state of this PR in my head).

By "cranking through it" I'm referring to the work of cleaning up this PR to a presentable state, making sure it passes CI, has appropriate documentation etc. Feel free to jump in, otherwise I'll get to it at some point, as stated.

davidszotten:
thanks for clarifying! will see if i can help fix any of the things ci is sad about

davidszotten:
hi again. i think this branch is exposing some other bugs in the codebase (that accidentally passed the test suite before but isn't with the stricter behaviour around nullable/not nullable currencies.)

would you prefer to discuss those here? or as separate bugs/prs?

davidszotten:
ok i think this is the main one #657

davidszotten:
rest of my test fixes (2 toods left i think): https://github.com/django-money/django-money/compare/main...davidszotten:py-moneyed-2b

jayvdb:
openSUSE is now using this patch @ https://build.opensuse.org/package/show/openSUSE:Factory/python-django-money

setup.py
benjaoming: If we focus on pymoneyed, isn't the experience that we *should* in fact pin it below the next major version? Since we are talking django-money, this is a transient dependency that might make upgrade paths more difficult if it doesn't have a roof.

djmoney/money.py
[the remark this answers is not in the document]
benjaoming: I don't think the function was "botched", but the rewrite adds a bit of clarity to a segment of code that was hard to read :+1:

benjaoming:
@antonagestam great work!! do you have time to fix the conflicts? And would you think it's okay to move this out of draft mode, given that your patch is now in OpenSUSE? :) I'd be happy to merge and release this!

Any thoughts @Stranger6667 ?

antonagestam:
@jayvdb @benjaoming It puzzles me for many reasons that this is in an OS 🤔

@benjaoming I could probably make time to fix conflicts, but I'm a little worried about untested behavior, such as what happens with migrations of existing usages. If you feel confident that this is good enough to merge, I can go head and clean it up though.

tests/testapp/models.py
```diff
@@ -24,6 +24,7 @@
 
 
 class ModelWithVanillaMoneyField(models.Model):
+    # TODO: This should now be an error?
```
antonagestam, line 27: @benjaoming I think this TODO should be addressed.

Stranger6667: Could you, please, elaborate a bit?

Stranger6667:
@benjaoming @antonagestam 

I think we need to move forward with this patch. As far as I see, there should not be any big issues with these changes :)

tests/testapp/models.py
antonagestam: @Stranger6667 I think my original reasoning here is that it should be an error to provide a default value without a currency. Regardless of whether or not that makes sense, I should the comment should either be acted on or removed.

antonagestam:
If anyone has time to spend, please feel free to take over here 🙂

benjaoming:
We'll bump the version of django-money substantially before releasing this.

Regarding testing.. I don't know what significance it had that someone put this draft patch into OpenSUSE... but I could think that they tested it beforehand, so let's say that it just weighs in positively for the decision.

benjaoming:
@flaeppe do you have bandwidth to help wrap up this PR? I haven wrestled with the default values as much as you. Could you provide a final solution for this merge conflict? :pray: 

```python
<<<<<<< feature/py-moneyed-2
            default = Money(default.amount, default.currency)
        if default is not None and default is not NOT_PROVIDED and not isinstance(default, Money):
            raise ValueError(f"default value must be an instance of Money, is: {default}")
        return default
=======
            amount, currency = default.amount, default.currency
        else:
            raise ValueError(f"default value must be an instance of Money, is: {default}")

        assert currency is not None, (
            "Default currency can not be `None` when default is of a value that doesn't include a currency. Either"
            " provide a default containing a currency value or configure a currency default via setting"
            " `default_currency=` or `DEFAULT_CURRENCY`(global)"
        )
        return Money(amount, currency)
>>>>>>> main
```

flaeppe:
@benjaoming Sure I'm having a look.

tests/testapp/models.py
flaeppe: This field now raises an error due to #661 being merged. i.e. this has to be corrected, it no longer makes sense to have a default value that can't be converted to a `Money` instance (-> needs a currency default or value that isn't `None`)

flaeppe: While refreshing my memory I think this is where I built #661 from, while attempting to move this PR forward. Though I chose to open a separate PR as it needed functional changes which weren't depending on changing the default currency to `None`.

I'm pretty what should happen here is to remove `default=0.0`

flaeppe:
@benjaoming I don't have permission to push any commits here directly. But I've resolved all that was red in the test suite. Can be found here: https://github.com/django-money/django-money/compare/main...flaeppe:feature/py-moneyed-2. Think you have access to push the commits I added directly on to here(?)

I've added 1 additional test there, which is breaking, intentionally. Basically to highlight that there's an issue in this PR that is still unresolved:

> ... 
I'd guess a natural replacement could be MoneyField(..., default_currency=None)(With that I mean via DEFAULT_CURRENCY = None). Without any other adjustments that would lead to a nullable CurrencyField and without using option MoneyField(..., null=True) the "default" setup would have a non-nullable money-column. And looking at this https://github.com/django-money/django-money/issues/530#issuecomment-558081994 regarding integrity options the "invalid" case is hit as default(which is controlled by MoneyField's descriptor class).
...

https://github.com/django-money/django-money/pull/638#issuecomment-1002479400

To clarify the issue; now, when using `MoneyField` without any defaults e.g.:

```python
class MoneyModel(models.Model):
    money = MoneyField(max_digits=10, decimal_places=2)
```

Will create the `money` column as __non nullable (null=False)__, while `money_currency` will be __nullable (null=True)__. IMO, as previously mentioned, I think we should change the lines below to not get a nullable currency per default

https://github.com/django-money/django-money/blob/a6867ea1b2c79bf46f10de1e251c77b97d0fe1f7/djmoney/models/fields.py#L265-L276

-> Change `null` argument to align with amount column being nullable, as default behaviour.

```python
         currency_field = CurrencyField( 
             ...,
             null=self.null, 
         ) 
```

tests/testapp/models.py
antonagestam: @flaeppe Your reasoning here sounds exactly right to me 👍

I added you as collaborator on my fork, so you should be able to push to this branch. But also feel free to replace this PR with a fresh one :)

flaeppe: Cool, thanks @antonagestam. I think I managed to merge `master` properly in to this branch.

flaeppe:
I think it'd be valuable if someone were to have a look again here that I've removed stuff related to `decimal_places_display` properly.

djmoney/settings.py
```diff
@@ -1,15 +1,14 @@
...
 from django.conf import settings
 
-from moneyed import CURRENCIES, DEFAULT_CURRENCY, DEFAULT_CURRENCY_CODE
+from moneyed import CURRENCIES, Currency
 
 
 # The default currency, you can define this in your project's settings module
 # This has to be a currency object imported from moneyed
-DEFAULT_CURRENCY = getattr(settings, "DEFAULT_CURRENCY", DEFAULT_CURRENCY)
+DEFAULT_CURRENCY: Currency = getattr(settings, "DEFAULT_CURRENCY", None)
```
antonagestam, line 11: Should we instantiate `Currency` here?

```suggestion
DEFAULT_CURRENCY = Currency(getattr(settings, "DEFAULT_CURRENCY", None))
```

(Not sure if `Currency` takes a `Currency` so might need to be more involved)

flaeppe:
I've gone through with the adjustments mentioned in https://github.com/django-money/django-money/pull/638#issuecomment-1146806167 in a8c5d60464f7ae68adb433eeeeacfabe657e1902. Let me know if there's a different, more preferable approach.

There's one (odd?) case that hasn't been resolved yet. Which is to have a nullable amount column with a non nullable currency column, though personally I can't figure out any use cases for that.

Although I have a suggestion for how to support it, if it feels relevant. Which would be to explicitly declare a `MoneyField` together with a `CurrencyField` to achieve that behaviour:

```python
class MoneyModel(models.Model):
    money = MoneyField(..., currency_field="money_currency", null=True)
    money_currency = CurrencyField(..., null=False)
```

But IMO that support should be pushed to a future PR.

tests/testapp/models.py
flaeppe: Should be possible to resolve this comment. The `TODO` is gone together with `default=0.0` and affected tests have been adjusted.

antonagestam:
@flaeppe Nice 👍 I have no objections to merging this PR as-is.

djmoney/settings.py
flaeppe: Nice catch 👍 , just tried passing a `Currency` instance. Didn't work. `DEFAULT_CURRENCY` has to be a `str`.

flaeppe: Although, think we need to handle `None` here as well.

flaeppe: Just realised the comment says "This has to be a currency object imported from moneyed". So, perhaps we shouldn't do anything new here?

jayvdb:
@Irostovsky , please create a new issue, or ask questions on stackoverflow,etc

PR #647, opened 2021-12-27
Include default_currency in deconstruct when not matching __init__ value

### LINKED ISSUE #646: default_currency=None keeps detecting migration changes
Based on the comment in the discussion found here: https://github.com/django-money/django-money/issues/530#issuecomment-650166087

If `MoneyField(..., default_currency=None)` is added to an already existing model, so that the field is triggering migration operations `AddField` or `AlterField`. Django keeps detecting changes on the field/model when running `python manage.py makemigrations`.

flaeppe:
Fixes #646 

---

Since `MoneyField.__init__` specifies `default_currency=DEFAULT_CURRENCY`, we should include `default_currency` in deconstruct-kwargs appropriately as soon as `self.default_currency != DEFAULT_CURRENCY`.

I think the changes here aligns with the documentation:

> If you haven’t added any extra options on top of the field you inherited from, then there’s no need to write a new deconstruct() method. __If, however, you’re changing the arguments passed in \_\_init\_\_() (like we are in HandField), you’ll need to supplement the values being passed__.

More on Django's field deconstruction can be found here: https://docs.djangoproject.com/en/dev/howto/custom-model-fields/#field-deconstruction

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -298,8 +298,11 @@ class MoneyField(models.DecimalField):
 
         if self._has_default:
             kwargs["default"] = self.default.amount
-        if self.default_currency is not None and self.default_currency != DEFAULT_CURRENCY:
-            kwargs["default_currency"] = str(self.default_currency)
+        if self.default_currency != DEFAULT_CURRENCY:
+            if self.default_currency is not None:
+                kwargs["default_currency"] = str(self.default_currency)
+            else:
+                kwargs["default_currency"] = None
         if self.currency_choices != CURRENCY_CHOICES:
             kwargs["currency_choices"] = self.currency_choices
         if self.currency_field_name:
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -787,3 +787,11 @@ def test_mixer_blend():
         instance = mixer.blend(ModelWithTwoMoneyFields)
         assert isinstance(instance.amount1, Money)
         assert isinstance(instance.amount2, Money)
+
+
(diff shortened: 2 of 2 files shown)


flaeppe:
With a quick peek through the code I couldn't find any other tests in regards to `MoneyField.deconstruct`. Think there could be a couple of more cases one would want to have coverage for?

flaeppe:
Oh, and I completely missed out on #532 ....

djmoney/models/fields.py
```diff
@@ -298,8 +298,8 @@ def deconstruct(self):
 
         if self._has_default:
             kwargs["default"] = self.default.amount
-        if self.default_currency is not None and self.default_currency != DEFAULT_CURRENCY:
-            kwargs["default_currency"] = str(self.default_currency)
+        if self.default_currency != DEFAULT_CURRENCY:
+            kwargs["default_currency"] = str(self.default_currency) if self.default_currency is not None else None
```
Stranger6667, line 302: I think it would be nice to have this as separate statements, so coverage info is immediately visible

jrocketfingers:
> Oh, and I completely missed out on #532 ....

No worries, I'm just glad that someone picked up the torch 🎉

djmoney/models/fields.py
flaeppe: Sure, I can fix that

flaeppe: @Stranger6667 I unwrapped the oneliner into a multi-line if statement at the same place. Let me know if you thought of it differently

Stranger6667 (approved):

PR #653, opened 2022-01-03
Drop unsupported Django and Python versions

stianjensen:
Django 3.1 is unsupported as of December 7, 2021
https://www.djangoproject.com/download/

Python 3.6 is unsupported as of December 23, 2021
https://endoflife.date/python

This replaces them in the test suite by Django 4.0 and Python 3.10


benjaoming (approved):
Thanks @stianjensen for this, it makes total sense to drop two unsupported old test matrix entries and replace them with two new supported/upcoming versions.

stianjensen:
I updated version numbers to be string now, since yaml interpreted 3.10 as 3.1 incorrectly.

stianjensen:
Looks like pypy-3.6 doesn't work with Django 4 or something. Tried updating CI to use pypy-3.8 instead.

stianjensen:
I tried handling test failures from django main, but having issues figuring out exactly what is changed that causes the test `test_rename_field` to fail.
I see that the messages for renamed fields are changed, but cannot figure out why that was not a part of the test previously:
https://github.com/django/django/commit/02bc7161ec477afd4a7b328936eb8adac078d7b9#diff-e7df880bdc17c719e0332fa0cfbd4eff49bd481f638e34335a6311cfcd0ebc26

stianjensen:
Also, the `rstcheck` part of pre-commit seems to fail due to something I did in the changelog. Not sure what to do with that 🤔

benjaoming:
@stianjensen the issue is in `docs/changes.rst` - I had to comment out an unused link reference for pre-commit to work earlier, but comment back in this line, and it should work:

( remove `#` )

```
# .. _Unreleased: https:///github.com/django-money/django-money/compare/2.1.1...HEAD
```

I'm sorry about this, we could perhaps try to retain the "Unreleased" section systematically straight after the next release.. or figure out a way to relax this pre-commit check.

stianjensen:
A-ha, now I see! Not that familiar with .rsc so I didn't get that's what the underscore meant!

Should be all good now, then :+1:

benjaoming:
> Not that familiar with .rst so I didn't get that's what the underscore meant!

That's totally fair -- it's not a great barrier for contributions to force people to read and fix reST syntax in order to contribute :see_no_evil:

Stranger6667 (approved):
Awesome! Thank you @stianjensen @benjaoming :heart:

benjaoming:
I tried to resolve the conflicts, it seems to look good?

stianjensen:
> I tried to resolve the conflicts, it seems to look good?

Yup, seems to have applied nicely :+1:

tests/test_models.py
```diff
@@ -724,7 +724,7 @@ def test_properties_access():
     elif VERSION[:2] > (2, 1):
```
stianjensen, line 724: This elif seems to have become redundant now, but that can also be cleaned up later. The else branch can be deleted and the elif turned into an else, since we don't test anything below django 2.2.

I guess the issues with django main were fixed in another PR as well.

benjaoming: Sounds good to drop it later, it seems to not be part of the PR changeset, so I think we can defer here.

Issue #654, opened 2022-01-06
MoneyField raises KeyError when model fetched via only()


danmoz:
As far as I can tell, MoneyField isn't compatible with querysets which use [only()](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.only)

To reproduce, fetch a model instance from a queryset which uses only(), and then evaluate that field; a KeyError is returned:

```
>>> invoice = Invoice.objects.filter(pk=1).only("amount").first()
>>> invoice.amount
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/venv/lib/python3.9/site-packages/djmoney/models/fields.py", line 103, in __get__
    data[self.field.name] = self._money_from_obj(obj)
  File "/venv/lib/python3.9/site-packages/djmoney/models/fields.py", line 91, in _money_from_obj
    currency = obj.__dict__[self.currency_field_name]
KeyError: 'amount_currency'
```

Tested on v2.1.1.

benjaoming:
Hi! Thanks for reporting this and creating a proof of concept test. What happens if you run with `.only("amount", "amount_currency")` ?

danmoz:
Yes, that works:

```
>>> invoice = Invoice.objects.filter(pk=1).only("amount", "amount_currency").first()
>>> invoice.amount
<Money: 0.00 AUD>
```

It's definitely a reasonable workaround. Ideally though, it would automagically work with just the model field name. I suppose an alternative solution would be to raise an exception if `only()` is detected in a `MoneyField` queryset, guiding the user to the workaround.

benjaoming:
In theory, one could want to *only* select an `amount`, for instance given that the currency is already filtered. `MyModel.objects.filter(amount_currency='USD').only('amount')` would make sense -- but will fail.

So I think that it could be argued that it's the string representation that assumes `amount_currency` field -- and `amount_currency` should not be assumed when the selected field sets are filtered.

I haven't worked with this area before, but perhaps you can fix the issue by making a check on whether a model instance has deferred fields? See: https://pandafy.me/deferred-fields-in-django

danmoz:
I think that is a valid case, but also very much an edge case.

In the example above, what is the value `instance.amount_currency` after line 1 and before line 2... `None`? I suspect that setting the currency to a temporarily undefined state just to support this edge case could cause more issues than it solves, don't you think?

PR #655, opened 2022-01-06
xfail test to demonstrate lack of support for only() querysets

danmoz:
Adding an xfail test for issue #654  to demonstrate how django-money throws a KeyError when used with only()


benjaoming (approved):
Agreeing with the xfail test case as a great measure, thanks for that! Apologies for the late processing :)

PR #657, opened 2022-01-19
Set value check

davidszotten:
suggested better fix for #427 

TODO: is the currency field guaranteed to appear before the amount (and main)
field? yes because of the creation_counters?

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -104,7 +104,12 @@ class MoneyFieldProxy:
         return data[self.field.name]
 
     def __set__(self, obj, value):  # noqa
-        if value is not None and self.field._currency_field.null and not isinstance(value, MONEY_CLASSES + (Decimal,)):
+        if (
+            value is not None
+            and self.field._currency_field.null
+            and not isinstance(value, MONEY_CLASSES)
+            and not obj.__dict__[self.currency_field_name]
+        ):
             # For nullable fields we need either both NULL amount and currency or both NOT NULL
             raise ValueError("Missing currency value")
         if isinstance(value, BaseExpression):
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+`Unreleased`_ - TBA
+-------------------
+
+**Fixed**
+
+- Improve the internal check for whether a currency is provided `#657`_ (`davidszotten`_)
+- Fix test suite for django main branch `#657`_ (`davidszotten`_)
+
 `2.1.1`_ - 2022-01-02
 ---------------------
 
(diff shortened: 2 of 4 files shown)


benjaoming (approved):
I agree that it shouldn't be possible to set a MoneyField value to simply a Decimal w/o currency :+1:

benjaoming:
@davidszotten is it okay that I converted this from Draft mode? I think that the changes look reasonable and they don't seem to trigger the former regression test, so everything should be fine?

Stranger6667:
Oh, maybe one thing - a changelog entry :)

Stranger6667:
Something about disallowing passing plain `Decimal` to the `MoneyField` should be fine :)

davidszotten:
> Something about disallowing passing plain `Decimal` to the `MoneyField` should be fine :)

that's not what we're changing... we are improving (i hope) the internal field check for whether a currency was provided

this is all a bit hacky because we (ab)use the same field for both the "public" money instance _and the internal `amount` value. it _might have been easier to do these checks if _two separate/internal columns were added (for amount as well as the current one for currency) but i guess that ship sailed long ago

Stranger6667:
> that's not what we're changing... we are improving (i hope) the internal field check for whether a currency was provided

Agree, it is indeed better to use the actual problem description, than a concrete example I was thinking of.

> this is all a bit hacky because we (ab)use the same field for both the "public" money instance _and the internal amount value. it _might have been easier to do these checks if _two separate/internal columns were added (for amount as well as the current one for currency) but i guess that ship sailed long ago

Yeah, it somehow reminds me about my [Building a custom model field from the ground up](https://www.youtube.com/watch?v=g12fkl9NCrM) talk at DjangoCon Europe 2019, where I prototyped an alternative `MoneyField`, that was actually backed by SQL structured types and felt much more convenient implementation-wise (though it was a PoC). I would not mind if somebody would like to take more steps in this direction. However, it is still on the idea stage and there are a lot of unknowns - but, I'd be happy to discuss it :)

davidszotten:
changelog entry added. does that look ok?

Stranger6667:
Looks good to me!

benjaoming:
Suggestion for next step: I think that this should be a 2.2 release because the behavior *is* different on an API level, and it can break a real-world scenario (where people somehow depend on being able to set the Decimal firstly, leave the currency unknown and set it afterwards) ?

davidszotten:
in that case maybe we should just fold this into the py-moneyed-2.0 = django-money-3.0 work?

benjaoming:
It could be released on its own, but if you're okay with waiting then it's fine for me :)

davidszotten:
sounds good. my only interest in this is in the context of py-moneyed-2.0. thanks!

Issue #621, opened 2021-05-23
Support py-moneyed 2.0


antonagestam:
While py-moneyed 2.0 hasn't been released yet, I thought it was worth keeping an issue open with the plan laid out in #567, copying it in here:

#### In version 2.0 [done]

- The py-moneyed dependency is pinned to `>1,<2`.
- Setting and reading `decimal_places_display` starts giving a deprecation warning.
- Setting `CURRENCY_DECIMAL_PLACES_DISPLAY` in settings starts giving a
  deprecation warning.
- `CURRENCY_DECIMAL_PLACES_DISPLAY` and `Money.decimal_places_display`
  both being unset results in `Money.__str__` using the new `l10n.format_money()`.
- Settings and properties for `format` and `currency_digits` are introduced,
  similar to how `decimal_places_display` currently works.

#### In version 3.0

- The py-moneyed dependency is pinned to `>2`.
- Calling the old `localization.format_money()` is removed, we are now always
  calling the new `l10n.format_money()`.
- The `decimal_places_display` property is removed.
- The `CURRENCY_DECIMAL_PLACES_DISPLAY` setting is removed.

Stranger6667:
`py-moneyed==2.0` is released :tada:

SchrodingersGat:
Note that unless you specify a particular version of py-moneyed (e.g. in requirements.txt) then django-money will now break:

![image](https://user-images.githubusercontent.com/10080325/119759681-0297e380-beec-11eb-93d4-f05aec78241c.png)

antonagestam:
@SchrodingersGat That should only be true if py-moneyed is a first-hand dependency. django-money's py-moneyed dependency is already pinned to <2.

SchrodingersGat:
My project is pinned to v1.1 which requires py-moneyed >=0.8

(I realize that 1.1 is a very old release now!)

antonagestam:
@SchrodingersGat Ah OK, then yes you'll have to pin it. Note that that is usually how Python dependencies are specified in packages, e.g. no upper bounds. See for instance [Django's dependencies](https://github.com/django/django/blob/main/setup.cfg#L39).e

Also note that it's best practice for applications to use a tool like pip-tools to pin all project dependencies to specific versions, not just first-hand dependencies.

antonagestam:
Just an FYI. Before working on this ticket I intend to first:

- Work on dropping support for legacy versions of Django and Python https://github.com/django-money/django-money/pull/612.
[...]

PR #661, opened 2022-03-08
Better default setup on `MoneyField` for models

flaeppe:
Hi, I was digging in to making some progress on #638 (refs: #621) and realised that there was some adjustments that could be made for defaults on `djmoney.models.fields.MoneyField`. I've gathered several (minor) fixes in this PR(see list below), all of them originating from the same method.

- Removes possibility of allowing _any_ currency code on string based default value
- Having a default as type `bytes` is now properly supported
- Asserts on incompatible defaults with default currency being `None` when amount has a valid type (i.e. a default setup generating `Money(10, None)`)

_I think_ the changes here pushes work for #638 at least a little bit forward, as one of the major changes there would (probably?) be setting `DEFAULT_CURRENCY=None` per default. While the changes here help out on management of `MoneyField(..., default_currency=None)`. And I was thinking it could be nice taking the opportunity helping reduce the size of the work on #638.

Anyways, let me know what you think.

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -91,6 +91,8 @@ class MoneyFieldProxy:
         currency = obj.__dict__[self.currency_field_name]
         if amount is None:
             return None
+        elif currency is None:
+            raise TypeError("Currency code can't be None")
         return Money(amount=amount, currency=currency, decimal_places=self.field.decimal_places)
 
     def __get__(self, obj, type=None):
@@ -182,7 +184,7 @@ class MoneyField(models.DecimalField):
     ):
         nullable = kwargs.get("null", False)
         default = self.setup_default(default, default_currency, nullable)
-        if not default_currency and default is not NOT_PROVIDED:
+        if not default_currency and default not in (None, NOT_PROVIDED):
             default_currency = default.currency
 
         self.currency_max_length = currency_max_length
@@ -196,25 +198,34 @@ class MoneyField(models.DecimalField):
         Field.creation_counter += 1
 
     def setup_default(self, default, default_currency, nullable):
-        if isinstance(default, (str, bytes)):
+        if default in (None, NOT_PROVIDED) or isinstance(default, Money):
+            return default
+        elif isinstance(default, (str, bytes)):
             try:
                 # handle scenario where default is formatted like:
                 # 'amount currency-code'
-                amount, currency = default.split(" ")
+                amount, currency = (default.decode() if isinstance(default, bytes) else default).split(" ", 1)
             except ValueError:
                 # value error would be risen if the default is
                 # without the currency part, i.e
                 # 'amount'
-                amount = default
+                amount = default.decode() if isinstance(default, bytes) else default
                 currency = default_currency
-            default = Money(Decimal(amount), Currency(code=currency))
+
+            amount = Decimal(amount)
         elif isinstance(default, (float, Decimal, int)):
-            default = Money(default, default_currency)
+            amount, currency = default, default_currency
         elif isinstance(default, OldMoney):
-            default = Money(default.amount, default.currency)
-        if default is not None and default is not NOT_PROVIDED and not isinstance(default, Money):
-            raise ValueError("default value must be an instance of Money, is: %s" % default)
-        return default
+            amount, currency = default.amount, default.currency
+        else:
+            raise ValueError(f"default value must be an instance of Money, is: {default}")
+
+        assert currency is not None, (
+            "Default currency can not be `None` when default is of a value that doesn't include a currency. Either"
+            " provide a default containing a currency value or configure a currency default via setting"
+            " `default_currency=` or `DEFAULT_CURRENCY`(global)"
+        )
+        return Money(amount, currency)
 
     def to_python(self, value):
         if isinstance(value, MONEY_CLASSES):
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -5,7 +5,7 @@ Created on May 7, 2011
 """
 import datetime
 from copy import copy
-from decimal import Decimal
+from decimal import Decimal, InvalidOperation
 
 from django import VERSION
 from django.core.exceptions import ValidationError
@@ -19,6 +19,7 @@ import pytest
 from djmoney.models.fields import MoneyField
 from djmoney.money import Money
 from moneyed import Money as OldMoney
+from moneyed.classes import CurrencyDoesNotExist
 
 from .testapp.models import (
     AbstractModel,
@@ -41,6 +42,7 @@ from .testapp.models import (
     ModelWithDefaultAsStringWithCurrency,
     ModelWithNonMoneyField,
     ModelWithNullableCurrency,
+    ModelWithNullDefaultOnNonNullableField,
     ModelWithSharedCurrency,
     ModelWithTwoMoneyFields,
     ModelWithUniqueIdAndCurrency,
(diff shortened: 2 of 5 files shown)


djmoney/models/fields.py
```diff
@@ -196,25 +196,34 @@ def __init__(
         Field.creation_counter += 1
 
     def setup_default(self, default, default_currency, nullable):
-        if isinstance(default, (str, bytes)):
+        if default in (None, NOT_PROVIDED) or isinstance(default, Money):
```
Stranger6667, line 199: Should it be `MONEY_CLASSES` in `isinstance` as in other places above?

flaeppe: I don't think so, as a bit further down we'll convert `OldMoney -> Money`

https://github.com/flaeppe/django-money/blob/e7d558c9cdf92227e997ef65fb6f1d3cdf87c14b/djmoney/models/fields.py#L216-L217

flaeppe:
I'm not sure why CI runs with `pypy3` now has an sqlite version that's incompatible with `django`.

flaeppe:
> I'm not sure why CI runs with `pypy3` now has an sqlite version that's incompatible with `django`.

I think the closest I got to is somewhere here: https://stackoverflow.com/questions/65476852/make-pypy-in-virtualenv-use-shared-library-from-os-rather-than-its-own-copy-lib

No idea if deleting PyPy's copies of `libsqlite3.so` would actually help though..

flaeppe:
@Stranger6667 any insight on why running with `pypy` would fail on the sqlite version? Do you know if we could do something to resolve it?

Additionally, there's an odd error (new to me at least) on the `pre-commit` job. Can be seen here: https://github.com/django-money/django-money/runs/5633124684?check_suite_focus=true . Not sure if rerunning the workflow would avoid it

flaeppe:
> @Stranger6667 any insight on why running with `pypy` would fail on the sqlite version? Do you know if we could do something to resolve it?
> 
> Additionally, there's an odd error (new to me at least) on the `pre-commit` job. Can be seen here: https://github.com/django-money/django-money/runs/5633124684?check_suite_focus=true . Not sure if rerunning the workflow would avoid it

Seems that it was fixed with `actions/setup-python` action bumping pypy version for Ubuntu20 (https://github.com/actions/virtual-environments/commit/1ede8085f53bf7b38163389282bed3b8173b4d62)

benjaoming (approved):
Solid work :100: 

If @Stranger6667 isn't around for another review, I'll get it in within the next days - meanwhile, release notes for this are more than welcome :)

tests/test_models.py
```diff
@@ -302,6 +304,80 @@ def test_null_default(self):
         instance = NullMoneyFieldModel.objects.create()
         assert instance.field is None
 
+    def test_implicit_currency_field_not_nullable_when_money_field_not_nullable(self):
+        instance = ModelWithVanillaMoneyField(money=10, money_currency=None)
+        with pytest.raises(CurrencyDoesNotExist, match=r"No currency with code NONE"):
```
Stranger6667, line 309: I think it would be better to avoid implicit conversion to string and provide a more specific error message for `None`.

Stranger6667 (commented):
Thanks for pinging me @benjaoming :)

I have a small comment on one of the error messages

tests/test_models.py
flaeppe: I think I see what you're getting at, if my 5 cents are worth anything, I'm of the opinion that's within the domain of `py-moneyed`(?). These are the lines where `"NONE”` would originate from:

https://github.com/py-moneyed/py-moneyed/blob/c495642244f629e4ed330b25438edbd517175146/src/moneyed/classes.py#L193-L197

`None` would get cast to `"NONE"` before it's validated as a currency code.

In addition to that, if we look at the `__init__` signature of `class Money` in `py-moneyed` (that runs the linked code above) it doesn't define any support for passing `currency=None`;

https://github.com/py-moneyed/py-moneyed/blob/c495642244f629e4ed330b25438edbd517175146/src/moneyed/classes.py#L173-L179

https://github.com/py-moneyed/py-moneyed/blob/c495642244f629e4ed330b25438edbd517175146/src/moneyed/classes.py#L186-L189

As such, I'm leaning towards the interpretation that `None` _is already an_ unsupported value for a currency code, therefore resulting in undefined behaviour (although kindly enough with the version of `py-moneyed` that's used we get an error with `"NONE"` attempted as currency code).

---

I'm aware of that I linked from `HEAD` at `py-moneyed`, but even if we're using an older version of `py-moneyed` which defines this differently. I kind of don't think it's worth to duplicate the code to handle `None`:

https://github.com/py-moneyed/py-moneyed/blob/c495642244f629e4ed330b25438edbd517175146/src/moneyed/classes.py#L186-L189

But _if_ we should do something different for `None`, I think we should "backport" those lines, throwing a `TypeError`.

flaeppe: I also can't remember why I added this test in the first place. Might have been just for coverage or so..

benjaoming: Isn't the AssertionError raised in L221 what would be the expected outcome of this test case?

djmoney/models/fields.py
```diff
@@ -196,25 +196,34 @@ def __init__(
...
-        return default
+            amount, currency = default.amount, default.currency
+        else:
+            raise ValueError(f"default value must be an instance of Money, is: {default}")
+
+        assert currency is not None, (
+            "Default currency can not be `None` when default is of a value that doesn't include a currency. Either"
+            " provide a default containing a currency value or configure a currency default via setting"
+            " `default_currency=` or `DEFAULT_CURRENCY`(global)"
+        )
```
benjaoming, line 225: Is this part reachable in tests? And should it perhaps also figure as a ValueError?

flaeppe: Yeah, it's reachable. It's covered by these tests:

https://github.com/django-money/django-money/blob/6dcdbeb685731ebbcf2d38638b3f6f87e21e6f1e/tests/test_models.py#L316-L320

My idea behind raising an `AssertionError` was to signal that it's an exception that's not intended to be caught by the caller.

tests/test_models.py
flaeppe: Not really, since that line is for initiating a `MoneyField` on a model(default value for currency), while this is setting a value for an already initiated `MoneyField`.

flaeppe: > I think it would be better to avoid implicit conversion to string and provide a more specific error message for `None`.

I don't have too strong opinions handling `None` in django-money and raising a `TypeError`, if that's what you guys think is best.

Let me know and I'll bump the PR.

djmoney/models/fields.py
flaeppe: > My idea behind raising an AssertionError was to signal that it's an exception that's not intended to be caught by the caller.

While thinking twice about that, it'd be quite weird for someone to actually catch that exception anyways:

```python
class MyModel(models.Model):
    try:
        money = MoneyField(...)
    except ValueError:
        # ??
```

If you'd prefer a `ValueError` there, I'm fine with changing it.

benjaoming: I understand your great explanation and rationale -- and no, this is not at all important to entertain going deeper into!

benjaoming: Thanks for explaining!

tests/test_models.py
benjaoming: I'm happy that wrong input (perhaps an edge case) is blocked in a nice way that will guide the developer to solve the issue, so I think we can move on :+1:

flaeppe: I changed it to raise a `TypeError` just now.

benjaoming: @flaeppe - thanks for updating the solution!

@Stranger6667 - to me, this looks good and fine, does the elaboration and the update work for you?

Stranger6667: Sounds good to me!

Stranger6667: Thanks for all the details :)

benjaoming:
Thanks for the patience @flaeppe :pray: :hourglass_flowing_sand:  :rocket:

PR #662, opened 2022-03-09
Repair CI

flaeppe:
~The original issue is with sqlite versions. See error from first run: https://github.com/django-money/django-money/runs/5476001835?check_suite_focus=true~ _Seems to be fixed since GitHub actions runs with `pypy3.8.13-7.3.9` (https://github.com/actions/virtual-environments/commit/1ede8085f53bf7b38163389282bed3b8173b4d62)_

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.7'
+          python-version: '3.9'
 
       - run: pip install pre-commit
       - run: pre-commit run --all-files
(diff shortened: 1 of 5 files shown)


flaeppe:
Thought fails with PyPy3 could be resolved by attempting: https://stackoverflow.com/questions/65476852/make-pypy-in-virtualenv-use-shared-library-from-os-rather-than-its-own-copy-lib

Did not seem to fix sqlite versioning problems and slowed down CI quite significantly. Although `django 2.2` became green.

I'm kind of out of ideas.

flaeppe:
@Stranger6667 I think the changes in here should fix CI. Not the change in https://github.com/django-money/django-money/pull/662/commits/46ceeae53b2da7ec9389b1d3906f42df307671ae, where coverage report will be skipped on failing test suite runs.

Stranger6667:
Oh, nice! Thank you @flaeppe

PR #663, opened 2022-03-23
add CurrencyField for Django Rest Framework

nmdevanddesign:
- Introduces the corresponding DRF Field: CurrencyField
- modifes register_money_field into register_drf_fields

### DIFF
--- a/djmoney/apps.py
+++ b/djmoney/apps.py
@@ -7,9 +7,9 @@ class MoneyConfig(AppConfig):
 
     def ready(self):
         try:
-            from .contrib.django_rest_framework import register_money_field
+            from .contrib.django_rest_framework import register_drf_fields
 
-            register_money_field()
+            register_drf_fields()
         except ImportError:
             pass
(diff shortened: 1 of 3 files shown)


benjaoming (changes requested):
Hey! Sorry that the PR has been stuck for so long. I'm one of the maintainers, but I don't use DRF. I believe that there isn't anything problematic here, the former behavior is retained, and only new behavior is added.

It should have a test case, though? The recipe could be to 1) add a new Model with a CurrencyField in `tests/testapp.models` and 2) add a simple test case to `tests/contrib/test_django_rest_framework` borrowing from the patterns already implemented..

https://github.com/django-money/django-money/blob/main/tests/testapp/models.py
https://github.com/django-money/django-money/blob/main/tests/contrib/test_django_rest_framework.py

Oh and thanks for the contribution! Unless you beat me to it, I'll add the changelog entry. @Stranger6667 LMK if you have any concerns here.

PR #665, opened 2022-04-20
fix: attribute error in BaseMoneyValidator

kristina-solovyova:
`AttributeError: 'decimal.Decimal' object has no attribute 'amount'`

### DIFF
--- a/djmoney/models/validators.py
+++ b/djmoney/models/validators.py
@@ -26,8 +26,6 @@ class BaseMoneyValidator(BaseValidator):
         limit_value = self.get_limit_value(cleaned)
         if limit_value is None:
             return
-        if isinstance(limit_value, (int, Decimal)):
-            cleaned = cleaned.amount
         params = {"limit_value": limit_value, "show_value": cleaned, "value": value}
         if self.compare(cleaned, limit_value):
             raise ValidationError(self.message, code=self.code, params=params)


kristina-solovyova:
@benjaoming, could you look at this, please? 🙏

djmoney/models/validators.py
```diff
@@ -26,8 +26,6 @@ def __call__(self, value):
         limit_value = self.get_limit_value(cleaned)
         if limit_value is None:
             return
-        if isinstance(limit_value, (int, Decimal)):
-            cleaned = cleaned.amount
```
benjaoming, line 30: Please LMK how I might be misunderstanding this:

This oddly comment-free behavior seems to anticipate the following: If a limit has been specified as int or Decimal, then we should compare it to the `amount` property of the cleaned Money value (which is also int/Decimal).

This should make sense with the rest of the `__call__` method... if you look further down in L33, there is `self.compare(cleaned, limit_value)` which should only be possible if values are `int/Decimal, int/Decimal` and not `Money, int/Decimal`.

LMK if I'm misreading something here and please do say how this has happened and what code triggered it?

benjaoming: And thank you for contributing :pray:

phillipuniverse: @benjaoming this was happening to me, here's a live debugging session with actual values:

image

The problem is that `cleaned` is a Decimal.  This happens because of this `to_internal_value` result in a `Decimal` being returned:

image

Still investigating but fwiw I only saw this in my upgrade from 0.12.3 to 3.0.0.

phillipuniverse: @benjaoming ok I've got a very simple example. This seems to only happen with DRF.

```python

class MoneyPropertySerializer(serializers.Serializer):
    flat_fee = MoneyField(max_digits=10, decimal_places=2, validators=[MinMoneyValidator(0)])

class DemoMoneyPropertyFailureTest(TestCase):

    def test_markup_serializer(self):
        data = {
            "flat_fee": 1000.0,
        }
        serializer = MoneyPropertySerializer(data=data)
        serializer.is_valid(raise_exception=True)
```

It's a problem of the validator on the DRF side (Django 3.2, DRF 3.13.1). This results in the following exception:

```
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/shipwell-backend-core-cpwchxt0-py3.7/lib/python3.7/site-packages/rest_framework/serializers.py", line 483, in to_internal_value
    validated_value = field.run_validation(primitive_value)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/shipwell-backend-core-cpwchxt0-py3.7/lib/python3.7/site-packages/rest_framework/fields.py", line 569, in run_validation
    self.run_validators(value)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/shipwell-backend-core-cpwchxt0-py3.7/lib/python3.7/site-packages/rest_framework/fields.py", line 593, in run_validators
    validator(value)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/shipwell-backend-core-cpwchxt0-py3.7/lib/python3.7/site-packages/djmoney/models/validators.py", line 30, in __call__
    cleaned = cleaned.amount
AttributeError: 'decimal.Decimal' object has no attribute 'amount'
```

My guess is that you were attempting to use the validators in both the DRF and Django model context and it doesn't work in the DRF context, since the "cleaned" coming in is will always be a Decimal when deserializing.

phillipuniverse: More information - seems like this only an issue if you do not specify a currency value. If you include a currency value in the payload then the `limit_value` and the `cleaned` value are both `Money`.

hosamhamdy258: quick update to more clarify the issue as i debugged more in this issue

> In the serializer level certain data, like certain kwargs that you pass to the model-field get's discarded, and only specific kwargs are allowed/handled (since they are not always needed for that type of field, e.g. decimal_places for a string )

[from stackoverflow](https://stackoverflow.com/questions/30214551/django-rest-serializer-and-extra-attributes-from-custom-model-fields#comment48572192_30234121)

also added simple solution you can check on https://github.com/django-money/django-money/pull/722#issue-1831799016

benjaoming: Rather than deleting these lines, wouldn't it be better to check the type of `cleaned` before accessing the `cleaned.amount` property? It seems like the old behavior kind of wanted to do that but was making the wrong assumptions by checking the type of the `limit_value` instead?

hosamhamdy258: validators don't generate the issue but the value being passed to them is as MoneyField in this code below is _PrimitiveMoney works well but if condition evaluate to false returns amount as decimal i hope that trace would help to give you more explanation for it 

	    def  get_value(self, data):
	    
		    amount =  super().get_value(data)
		    
		    currency = data.get(get_currency_field_name(self.field_name), self.default_currency)
		    
		    if currency and amount is  not  None  and  not  isinstance(amount, MONEY_CLASSES) and amount is  not empty:
		    
		    return _PrimitiveMoney(amount=amount, currency=currency)
		    
		    return amount

benjaoming: I think that adding a test case as is shown in https://github.com/django-money/django-money/pull/665#discussion_r945319128 sounds great.

Maybe I'm not getting the full picture, but it sounds like everyone thinks that removing the former lines is a perfect fix?

In that case, it would be great if this is fixed in a way that _at least_ adds a comment to show exactly what is expected as types, not least for `limit_value`?

phillipuniverse:
Pretty sure this should be closed as superseded by #722.

benjaoming:
That's also how I read it :+1: 

Thanks for the contributions and investigations here!

PR #656, opened 2022-01-13
Fix Django 3.2 warning about default_app_config

karlwnw:
Fix the following warning:
```
django.utils.deprecation.RemovedInDjango41Warning: 'djmoney' defines
default_app_config = 'djmoney.apps.MoneyConfig'. Django now detects
this configuration automatically. You can remove default_app_config.
```

https://docs.djangoproject.com/en/4.0/releases/3.2/#automatic-appconfig-discovery
Copied from how django-extensions handles this:
https://github.com/django-extensions/django-extensions/blob/main/django_extensions/__init__.py

PR #667, opened 2022-05-16
Fix django app config default

huxaiphaer:
What does this PR do ?
- It solves the fix warnings `default_app_config`.

Full error details :

`RemovedInDjango41Warning: 'djmoney' defines default_app_config = 'djmoney.apps.MoneyConfig'. Django now detects this configuration automatically. You can remove default_app_config.`

### DIFF
--- a/djmoney/__init__.py
+++ b/djmoney/__init__.py
@@ -1,2 +1,10 @@
+import django
+
 __version__ = "2.1.1"
-default_app_config = "djmoney.apps.MoneyConfig"
+
+if django.VERSION >= (3, 2):
+    # The declaration is only needed for older Django versions
+    pass
+else:
+    default_app_config = "djmoney.apps.MoneyConfig"
+


djmoney/__init__.py
```diff
@@ -1,2 +1,9 @@
-__version__ = "2.1.1"
-default_app_config = "djmoney.apps.MoneyConfig"
+import django
+
+if django.VERSION >= (3, 2):
+    # The declaration is only needed for older Django versions
+    pass
+else:
+    __version__ = "2.1.1"
```
benjaoming, line 7: The `__version__` specification is for any version of Django. Will check up on other aspects. But this would have to be changed.

benjaoming: Just following up on this.. it seems we don't import anything during setup.py so I would think that it would be okay to import django here

benjaoming (approved):
No immediate concerns here. Would like to hear from others if they consider it problematic to import django directly in `__init__`?

djmoney/__init__.py
```diff
@@ -1,2 +1,10 @@
+import django
+
 __version__ = "2.1.1"
-default_app_config = "djmoney.apps.MoneyConfig"
+
+if django.VERSION >= (3, 2):
+    # The declaration is only needed for older Django versions
+    pass
+else:
+    default_app_config = "djmoney.apps.MoneyConfig"
```
benjaoming, line 9: Maybe all this can be a single ` if django.VERSION < (3, 2):` ?

benjaoming (changes requested):

benjaoming:
I didn't remember that #656 was already fixing this, thanks for the effort!

PR #679, opened 2022-07-13
Test automatic wrapping of model manager

stianjensen:
The docs say that any custom model manager will be automatically wrapped
with the money_manager (at least as long as it's named 'objects').
(https://github.com/django-money/django-money#important-note-on-model-managers)

I'm having some issue where it seems like that doesn't happen, though.
Trying to add a test case to show that it doesn't work as the
documentation explains it.

Am I misunderstanding something?


stianjensen:
> Huh – When I ran this test locally, it failed:

So it turns out the test passes if you run the whole test suite, but fails when running the test individually. I guess something stateful is kept between the tests which causes the behavior to change.

benjaoming:
@stianjensen could you try to merge in the latest `main`? :pray:

Issue #675, opened 2022-06-18
Getting 'HTTP Error 403: Forbidden' using Fixer Backend


drexhacker:
I already set the FIXER_ACCESS_KEY but I get `HTTP Error 403: Forbidden` after running `python manage.py update_rates`?

benjaoming:
Can you check if django-money is making a wrong API call? https://github.com/django-money/django-money/blob/main/djmoney/contrib/exchange/backends/fixer.py

benjaoming:
Why did you close the issue? Maybe the fixer backend needs an update? https://github.com/django-money/django-money/blob/main/djmoney/contrib/exchange/backends/fixer.py

drexhacker:
Yeah It needs an update

benjaoming:
Thanks for confirming! If anyone is interested in doing a PR, we'll get it shipped :+1:

drexhacker:
Hope to Contribute a lot to `django-money` When I get time.

JohnDoee:
It seems like fixer moved to apilayer. I can't create a new fixer account and I am redirected to the apilayer website on signup.

That would also align with it being a header instead of a query parameter as that's where their API seems to differ.

A new backend is probably warranted here to fix that.

benjaoming:
Thanks @JohnDoee ! This makes sense, a refactor of the current backend is most welcome :+1: I've updated the issue.

JohnDoee:
@benjaoming How do you feel about adding requests as a dependency? The current solution seems to be re-implementing a bit of an HTTP client wrapper that extends on demand and have caused more than one issue like #403.

With requests every backend would be 1 line of effective code and imo more robust.

I do understand if that's too much just to add easy header support though.

benjaoming:
I know that `requests` are very common. But I don't think we should add additional dependencies for an optional `contrib` application... UNLESS you can quickly convince me that `requests` is already contained somewhere in the dependency set of django-money :)

benjaoming:
I am 99% sure that the one who wrote the current implementation made the particular effort to avoid adding extra dependencies :)

spaut33:
It would be better to add support for both URL options (fixer.io and apilayer) and change the key name (`access_key` or `apikey`) depending on which URL is in settings. Because, in fact, this is not an obvious point.
_Originally posted by @ignat14 in https://github.com/django-money/django-money/pull/681#discussion_r1179340984_

fedosov:
[...]

PR #681, opened 2022-07-24
Update rates from fixer.io #675

spaut33:
Small fix for #675. Changed get parameter and added 'user-agent' header to avoid 403 error.

### DIFF
--- a/djmoney/contrib/exchange/backends/base.py
+++ b/djmoney/contrib/exchange/backends/base.py
@@ -2,7 +2,7 @@ import json
 import ssl
 from decimal import Decimal
 from urllib.parse import parse_qsl, urlparse, urlunparse
-from urllib.request import urlopen
+from urllib.request import Request, urlopen
 
 from django.db.transaction import atomic
 from django.utils.http import urlencode
@@ -45,9 +45,12 @@ class BaseExchangeBackend:
         return {}
 
     def get_response(self, **params):
-        url = self.get_url(**params)
+        headers = {
+            "User-Agent": "Mozilla/5.0",
+        }
+        request = Request(self.get_url(**params), headers=headers)
         context = ssl.create_default_context(cafile=certifi.where())
-        response = urlopen(url, context=context)
+        response = urlopen(request, context=context)
         return response.read()
 
     def parse_json(self, response):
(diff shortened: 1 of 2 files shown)


djmoney/contrib/exchange/backends/base.py
```diff
@@ -45,9 +45,12 @@ def get_params(self):
         return {}
 
     def get_response(self, **params):
-        url = self.get_url(**params)
+        headers = {
+            "User-Agent": "Mozilla/5.0",
```
benjaoming, line 49: You could get into trouble for faking the User-Agent, not sure. But I'll leave that up to others to scrutinize :)

benjaoming (approved):
LGTM :+1:

djmoney/contrib/exchange/backends/fixer.py
```diff
@@ -15,4 +15,4 @@ def __init__(self, url=settings.FIXER_URL, access_key=settings.FIXER_ACCESS_KEY)
         self.access_key = access_key
 
     def get_params(self):
-        return {"access_key": self.access_key}
+        return {"apikey": self.access_key}
```
ignat14, line 18: Why is this change added? Isn't the fixer api expecting an `access_key` parameter?
https://fixer.io/documentation

ignat14: Thanks @spaut33 for the quick response. It makes sense, but isn't the `apikey` suppose to be included in the headers in that case?
Sorry for the questions but a code I'm working on started failing after upgrading djmoney to 3.1.0 and I'm trying to figure out what's the issue.

spaut33: It works both ways, as a request parameter and as a header. However, I believe that APILayer may not be following the correct convention, as custom fields in headers should start with the "X-" prefix.

fedosov: Same here. Fixed by creating API key through ApiLayer and adding `FIXER_URL = "https://api.apilayer.com/fixer/latest"` to my `settings.py`

fedosov: But `apilayer` seems unreliable — getting timeouts accessing the API. Could we maybe send `access_key` and `apikey` params to suit both APIs?

spaut33: @ignat14 @fedosov Thank you for your comments. I updated the [issue](https://github.com/django-money/django-money/issues/675)

PR #685, opened 2022-08-16
Add app config for djmoney.contrib.exchange

### LINKED ISSUE #684: Migration created after makemigrations with DEFAULT_AUTO_FIELD modified settings
An migration called "0002_alter_rate_id.py" created after running "makemigrations" with
`DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'`

ventur40:
Fixes #684


benjaoming (changes requested):
Thanks, it needs to be called `apps.py` to work :)

benjaoming (approved):
LGTM! Thanks for fixing :rocket:

djmoney/contrib/exchange/apps.py
```diff
@@ -0,0 +1,3 @@
+class ExchangeConfig(AppConfig):
```
stianjensen, line 1: Looks like you're missing an import:
`from django.apps import AppConfig`

ventur40: Added!

alexnorgaard:
I can see that this is failing on some tests, but the logs are too old to view. Can it be re-run I can have a look? If it needs more  work, I will be happy to give it a go, as I need this fix for implementing django-money on my company's existing code base.

stianjensen:
Users with the correction permission should have access to a button like this on the top-right corner of the summary tab for checks:
<pre align="center"></pre>

benjaoming:
Doesn't exist and I should have all rights necessary :/

image

alexnorgaard:
> Doesn't exist and I should have all rights necessary :/
> 
> image

Not even if you click the Latest #2 button?

benjaoming:
Nope, it's a drop-down.. I've tried _everything_ afaict
image

stianjensen:
https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs

Aha, re-run is only available up to 30 days after the first run of a job.

benjaoming:
Nice observation @stianjensen ! It'd be nice of those UI people working at GitHub to just disable the button rather than completely removing it :)

alexnorgaard:
I have opened a new PR with the same changes here: https://github.com/django-money/django-money/pull/715 
The workflow needs to be approved

PR #687, opened 2022-08-31
add field generator for model bakery

PetrDlouhy:
Add field generator which ensures seamless model creation with [model bakery](https://github.com/model-bakers/model_bakery).

This doesn't create any additional dependency (except testing). It is used only if model_bakery is installed.

### DIFF
--- a/djmoney/__init__.py
+++ b/djmoney/__init__.py
@@ -8,3 +8,5 @@ try:
 except ModuleNotFoundError:
     # this part is useful for allow setup.py to be used for version checks
     pass
+
+from . import baker_field_generator
--- a/djmoney/baker_field_generator.py
+++ b/djmoney/baker_field_generator.py
@@ -1,0 +1,15 @@
+from decimal import Decimal
+
+from djmoney.money import Money
+
+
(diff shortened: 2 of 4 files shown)


benjaoming (commented):
I had a super-old review stuck. If you are still around @PetrDlouhy to update this, would you be able to find some different path for including model_bakery with a more generic purpose than just generating 100 USD? I'm not sure if having this as a general implementation "works for all" makes sense?

benjaoming:
Closing because of missing feedback, but always welcome to reopen :+1:

benjaoming:
@PetrDlouhy 

I think that the easiest fix for now is to take the recipe of this PR and turn it into a documentation section/chapter called "MoneyField test fixtures" :question: 

(then someone else can chip in with the Factory Boy parallel)

I would definitely merge a documentation contribution straight away :+1: 

(in fact, been wondering about if the next thing that should happen to this project, would be an overhaul and modernization on the documentation / tooling side, to reflect that in general this is a mature project, but it's missing some bits and pieces on the general impression)

PR #689, opened 2022-09-27
rate is always 1 if source and target are equal

### LINKED ISSUE #688: calling get_rate with identical source and target currency should be a noop
- it is always exactly 1
- that doesn't even have to hit the cache ;)

nschlemm:
we don't even need to cache that ;)

Fixes #688


benjaoming (approved):
LGTM :+1:

benjaoming:
Sorry, it seems there is a test case with `call_count` that's unhappy now. But should be straight-forward to fix that :)

nschlemm:
@benjaoming the mocking gods should be pleased now ;)

PR #686, opened 2022-08-20
Initial stub

karolyi:
Can we get some typing going? See the issue https://github.com/microsoft/pyright/issues/3848

PR #691, opened 2022-10-14
Add `py.typed` to start exporting types

sondrelg:
Adds an empty `py.typed` to the package, which will instruct static type checkers that this package supports static type checking of its code. See [PEP561](https://peps.python.org/pep-0561/), or the [typing docs](https://typing.readthedocs.io/en/latest/source/libraries.html#marking-a-package-as-providing-type-information) for more info.

(The pep states you should add the file in your setup.py `package_data` explicitly, but when testing without it, the file seems to be included in the wheel. Would be good to get a second opinion here, but I don't think that's necessary)

Then, the remaining commits adds mypy to ensure that added types appear correct. Since pre-commit is already run during CI, no changes to the pipeline should be necessary.


sondrelg:
For context, these were the initial mypy errors:

```
django-money on  main [!?] via 🐍 v3.12.0a0 (django-money) took 6s 
❯ mypy .
djmoney/settings.py:11: error: Incompatible types in assignment (expression has type "Optional[Any]", variable has type "Currency")
djmoney/contrib/exchange/migrations/0001_initial.py:13: error: Need type annotation for "dependencies" (hint: "dependencies: List[<type>] = ...")
djmoney/contrib/exchange/backends/openexchangerates.py:9: error: Incompatible types in assignment (expression has type "str", base class "BaseExchangeBackend" defined the type as "None")
djmoney/contrib/exchange/backends/openexchangerates.py:10: error: Incompatible types in assignment (expression has type "Union[Any, str]", base class "BaseExchangeBackend" defined the type as "None")
djmoney/contrib/exchange/backends/fixer.py:9: error: Incompatible types in assignment (expression has type "str", base class "BaseExchangeBackend" defined the type as "None")
tests/contrib/exchange/conftest.py:68: error: Incompatible types in assignment (expression has type "str", base class "BaseExchangeBackend" defined the type as "None")
tests/contrib/exchange/conftest.py:75: error: Incompatible types in assignment (expression has type "str", base class "BaseExchangeBackend" defined the type as "None")
tests/test_models.py:514: error: Unsupported operand types for % ("Money" and "F")
Found 8 errors in 6 files (checked 64 source files)
```

djmoney/contrib/exchange/backends/base.py
```diff
@@ -19,8 +20,8 @@
 
 
 class BaseExchangeBackend:
-    name = None
-    url = None
+    name: Optional[str] = None
+    url: Optional[str] = None
```
benjaoming, line 24: Okay I'm not sure about this, but according to Adam J's latest blog post, this might be old syntax? https://adamj.eu/tech/2022/10/18/python-type-hints-implicit-optional-types/

benjaoming (commented):
I think that this is going the right direction.

1) To what extend should type hints be found in the entire codebase? Is it sufficient to add it to everything that's assumed a public API?

2) Should all MyPy errors be resolved before it's ready?

djmoney/contrib/exchange/backends/base.py
Stranger6667: I believe, this syntax was introduced in Python 3.10, so likely we need to support the old one for a while

sondrelg:
> To what extend should type hints be found in the entire codebase? Is it sufficient to add it to everything that's assumed a public API?

My interpretation is that type checking is incremental. Untyped functions are interpreted to return `Any`. So you don't need a single annotation before adding `py.typed`, but the value from exported types increases with annotation coverage, since returning `Any` everywhere isn't particularly useful.

> Should all MyPy errors be resolved before it's ready?

They should be resolved here :+1:

benjaoming:
@sondrelg there were still some unresolved MyPy errors?

sondrelg:
No I don't think so. Where would they be? 🙂 Mypy was run in the pre-commit checks

benjaoming (approved):
LGTM :+1: 

@Stranger6667 ?

Stranger6667:
Looks good to me too! :)

Stranger6667 (approved):

karolyi:
Can you please add my changes from #686?

benjaoming:
@karolyi there is some test breakage, so if you can update the PR with the latest `main` and find a way for :heavy_check_mark:'s on all the tests, that'd be great :+1:

karolyi:
I can only do that with my PR, I'm afraid.

benjaoming:
That's what I meant, this one is merged :)

karolyi:
I need to spend a bit more time with that pre-commit failure. mypy is not my cup of tea so I don't know what its problem is; need to figure that out.

The interface file is proper.

benjaoming:
None of this is my current cop of tea, so I can't be of much help. But I'd be happy to push things along with the right bits of information and :heavy_check_mark:'s :)

PR #700, opened 2023-03-04
Add support for Python 3.11 and Django 4.2

### LINKED ISSUE #699: Python 3.11 and Django 4.2 support
Currently, django-money doesn't support Python 3.11. There shouldn't be any problems with the upgrade because library works fine under Python 3.10. Recently, Django dropped support for Python versions lower than 3.10. Because of that change, django-money can work only on Python version 3.10 for the next Django release. 

During working on this task, the job to add support for Django 4.2 should be similiar.

sdarmofal:
Python 3.11 - everything is OK.
Django 4.2 - it's a beta version. Working properly, excluding pypy3

I discovered that Django 4.1 and 4.2 doesn't work properly under pypy3. I mentioned this fact in the README. The error:

```
File "/home/runner/work/django-money/django-money/.tox/django42-pypy3/lib/pypy3.8/site-packages/django/contrib/admin/decorators.py", line 102, in _model_admin_wrapper

raise ValueError("site must subclass AdminSite")

ValueError: site must subclass AdminSite
```

Closes #699

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,10 +40,6 @@ jobs:
             tox_env: django32-pypy3
           - python: pypy-3.8
             tox_env: django40-pypy3
-          - python: pypy-3.8
-            tox_env: django41-pypy3
-          - python: pypy-3.8
-            tox_env: django42-pypy3
           - python: '3.7'
             tox_env: django22-py37
           - python: '3.7'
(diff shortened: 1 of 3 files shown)


benjaoming (approved):
Really good stuff :100: 

Thanks for updating all relevant places, including README. I agree that the pypy3 error mentioned looks like something we can't do much about. Not sure how many users this affects.. but they'll have to figure it out upstream or tell us if we were wrong.

PR #707, opened 2023-04-27
Fix: add support for both `data.fixer.io` and `api.apilayer.com` auth params

fedosov:
In the current state of the project, exchange rates request from fixer.io does not work. This change allows you to work with both available API entry points: `data.fixer.io` and `api.apilayer.com`.

ApiLayer seems unreliable to me, getting a lot of timeouts and HTTP Error 524.


benjaoming (approved):
Looks like a good way to support both. They'll get an extra empty parameter, but I guess that's the price of not agreeing to using the same name :man_shrugging:

PR #716, opened 2023-06-14
Explicitly set id field to AutoField on Rate model

alexnorgaard:


alexnorgaard:
I do not understand why this would give the following error on the django32-pypy3 tox job: `TypeError: Parameters to Generic[...] must all be type variables` . Anyone who have some insight?

benjaoming:
Maybe @karolyi or @sondrelg will know the answer, they have added some typing stuff lately.

alexnorgaard:
I have dug into the asgiref package, and it looks like the versions 3.7.1 =< are introducing parameterized Generic types using parameter expressions, one place in the code, which is introduced in python 3.10. Dunno why it is only breaking pypy3.8 though, and I don't think I am in a position to figure this out, but I have an inkling that this will fail on other future PR's as well. The change to asgiref was introduced in May.

alexnorgaard:
This issue https://github.com/django/asgiref/issues/393 on the asgiref repo explains the issue with pypy, and a suggestion of pypy typing incompatibility is made. Switching to asgiref 3.6 should make this work again.

benjaoming:
@alexnorgaard seems to work :+1: 

image

alexnorgaard:
Nice! How about the changes to the model?

Stranger6667:
Unfortunately, I don't have much input for this issue, but I think that a minor version bump + a proper release note should be enough for this, indeed :)

benjaoming:
> As the initial migration file for Rate is included in the project with an AutoField it can never be anything else.

Ah, that sounds great :+1: 

But just in case, I would definitely still mention something clever in the release notes.

Re:

> Alternatively, the migration file should dynamically set the id field to either BigAutoField or AutoField depending on the system settings, but that would be a large change.

This sounds like something to be avoided :)

benjaoming:
Ah, I mean these => https://github.com/django-money/django-money/blob/main/docs/changes.rst

Can you add the note to changes.rst in this PR?

alexnorgaard:
> Ah, I mean these => https://github.com/django-money/django-money/blob/main/docs/changes.rst
> 
> Can you add the note to changes.rst in this PR?

Done

docs/changes.rst
```diff
@@ -6,6 +6,18 @@ Changelog
 
 - TBC...
 
+`3.1.1`_ - 2023-06-15
+-------------------
+
+**Changed**
+
+- Explicitly define id in ``Rate`` model - This ensures that the database table will use ``AutoField`` even if ``DEFAULT_AUTO_FIELD`` is set to ``BigAutoField`` in the Django project's settings
```
benjaoming, line 14:
```suggestion
- Explicitly define ``id`` in ``djmoney.contrib.exchange.Rate`` model - This ensures that the database table will use ``AutoField`` even if ``DEFAULT_AUTO_FIELD`` is set to ``BigAutoField`` in the Django project's settings #716
```

benjaoming (approved):

djmoney/contrib/exchange/models.py
```diff
@@ -22,6 +22,7 @@ def clear_rates(self):
 
 
 class Rate(models.Model):
+    id = models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name="ID")
```
benjaoming, line 25: I haven't used the `serialize` option before, what does it do here?

alexnorgaard: The intention was to write out the id field just as it is defined in the 0001 migration, and as explicitly as possible. This way we're making sure that django will not try to make any new migrations because of a difference in the field.

alexnorgaard: The migration file defines the id field like this: `("id", models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name="ID")),`

benjaoming: :100: got it, thanks :+1:

alexnorgaard:
Sorry. Forgot to run pre-commit locally

alexnorgaard:
Can't make a referral to a tag that does not exist yet. Removed it in `changes.rst`

benjaoming:
Great work @alexnorgaard :100:

alexnorgaard:
Super! Do you know what the process for releasing 3.2 is?

benjaoming:
Made a note :)

alexnorgaard:
> Made a note :)

Sorry for pestering you about it, but does it mean that you made a note to remember to make a release or something else? I'm a little stuck with implementing djanog-money, until the changes are released :)

benjaoming:
Yes, I made a note to release it unless @Stranger6667 beats me to it.

alexnorgaard:
> Yes, I made a note to release it unless @Stranger6667 beats me to it.

Can I bug you one more time for the release of 3.2?

Stranger6667:
@alexnorgaard 

Done!

alexnorgaard:
> @alexnorgaard
> 
> Done!

Thank you so much!

benjaoming:
Hey, my TODO note finally sent me here looking to do a release - thanks for already doing that @Stranger6667 :tada:

PR #718, opened 2023-06-15
Tox: pypy3 downgrade asgiref

benjaoming:
@alexnorgaard this change probably fail (I'm sure I did some mistake in the tox config), but you could see if you can have something like this to fix the issue in https://github.com/django-money/django-money/pull/716#issuecomment-1592594627


alexnorgaard:
Alright. I'm not really familiar with Tox, and I'm not sure how all the dependencies are configured, as they are not in the tox.ini file, but I'll give it a go

Issue #725, opened 2023-09-01
Depending on field-order in earlier migrations, not all necessary migrations are detected


vanschelven:
In a specific combination of circumstances Django does not detect changes to the `CurrencyField` that's associated with a given `MoneyField` when it should. 

The problem occurs when the order in which the fields originally appear in your migrations is `MoneyField`-before-`CurrencyField`. As of `django-money==3.2` (and `Django==3.2`) this order occurs when a `MoneyField` is added, but not when it is originally present in an initial `CreateModel` migration.

Such a migration looks like so:

```
migrations.AddField(
    model_name='bar',
    name='price',
    field=djmoney.models.fields.MoneyField(decimal_places=2, default=Decimal('0.0'), default_currency='EUR', max_digits=14),
),
migrations.AddField(
    model_name='bar',
    name='price_currency',
    field=djmoney.models.fields.CurrencyField(choices=[('XUA', 'ADB Unit of Account'), ...], default='EUR', editable=False, max_length=3),
),
```

At a later point it may happen that the `CurrencyField` changes (but not the `MoneyField`), e.g. because of an upgrade to `pymoneyed`. This in turn affects the `choices` of the `CurrencyField` and should result in `makemigrations` detecting a necessary change. However, this necessity is not detected.

To understand why, note that, to detect whether changes in the model have been made, Django constructs a fake set of apps/models based on the existing migrations and compares this to the currently existing one.

However, when the fields are ordered Money-before-Currency, the addition of the MoneyField to the fake model in the fake set will [trigger the immediate addition of a `CurrencyField`](https://github.com/django-money/django-money/blob/aceac3ecc1f0202e0bc59cb712c8b11e00b422f7/djmoney/models/fields.py#L265) _of the current kind_, i.e. not as it was at the time of the migration, to the model. When it's finally time to add the `CurrencyField` as defined in the migration, [this is prevented by a check in the code](https://github.com/django-money/django-money/blob/aceac3ecc1f0202e0bc59cb712c8b11e00b422f7/djmoney/models/fields.py#L164). Later, when fake past and actual present are compared, they appear identical because the fake past is affected by the actual present and `makemigrations` declares there are no changes detected.

[...]

PR #726, opened 2023-09-01
Don't contribute a currency field when in the 'fake' migration world

vanschelven:
in that case the currency field should/will simply be added because it is explicitly defined as part of some migration.

See #725

### DIFF
--- a/djmoney/models/fields.py
+++ b/djmoney/models/fields.py
@@ -255,7 +255,7 @@ class MoneyField(models.DecimalField):
     def contribute_to_class(self, cls, name):
         cls._meta.has_money_field = True
 
-        if not hasattr(self, "_currency_field"):
+        if not hasattr(self, "_currency_field") and not cls.__module__ == "__fake__":
             self.add_currency_field(cls, name)
 
         super().contribute_to_class(cls, name)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,7 +4,9 @@ Changelog
 `Unreleased`_ (TBA)
 -------------------
 
-- TBC...
+**Fixed**
+
+- Fix detection of necessary migrations. Note that this means that previously undetected migrations will be detected as of this version  `#725`_ (`vanschelven`_)
 
 `3.2`_ - 2023-07-03
 -------------------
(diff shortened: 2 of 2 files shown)


benjaoming:
@vanschelven thanks so much for this impressive work! I think that your explanation makes sense and the change looks like the right one.

But I would just defer to @Stranger6667 for a comment here?

Would you mind adding a release note entry referencing the findings and explanation in #725 ? The bug and its fix is important for others to be aware of, as both the bug and the subsequent fix cause migrations to happen certain circumstances - right? :pray:

vanschelven:
I've added a note to the changelog. Your understanding is correct: because this is now fixed, upgrading to the version of django-money with this fix will start detecting migrations, which may in itself be surprising.

benjaoming (approved):
Thanks for adding release notes! I don't have anything to add here, will ask @Stranger6667 to have a look :hourglass_flowing_sand:

Stranger6667 (approved):
Looks good to me!

benjaoming:
Great! We'd better bump the release to 3.3 because of the significance of the change...

stianjensen:
Great stuff! I'm pretty sure this is what's been blocking us from upgrading from 2.1.1. Will test this to confirm our issue is also gone now :+1:

vanschelven:
If we're being "paranoid" (and given the lack of tests) I'd like to point out one more thing.

In [the code that I used as an example of how it used to work when "south" was still a thing](https://github.com/django-money/django-money/commit/69b51755f5bd18e4b85f9887aced69998d462e66#diff-0b1ce35a3277e63b0a1f1bd7c798dbc656d3ab504c188d6d9553095759caf800L283) there is one more if-statement like the one I added in this PR.

Given my analysis on the issue-description that if-statement seems redundant to me, and it did not appear to be necessary in the manual tests that I ran on my live project, but if we want to do "better safe than sorry" we could add it in as well.

stianjensen:
> Can you report back if that's the case? I'll put out the final release after the additional verification and encouragement

All good from me on this! We've been using the currency_field_name to share currency between multiple moneyfields and also adding a custom CharField to our models, and some of the hacks we've had on our end are now unnecessary, probably thanks to this fix.

vanschelven:
> Great! We'd better bump the release to 3.3 because of the significance of the change...

@benjaoming kind reminder to do this

benjaoming:
Released here: https://pypi.org/project/django-money/

Issue #636, opened 2021-09-29
Do not register custom json serializer implicitly instead provide registration information in the docs


emorozov:
Hello,
djmoney registers its own json deserializer for django-reversion in the `models/__init__.py`. This seems to be inconvenient, because:
a) such registration will override any custom serializers set by user in the `settings.py` (`SERIALIZATION_MODULES`) causing errors that are pretty difficult to debug
b) it is hard to override djmoney override because it is done at model import time, if I need my custom serializer for json, where would I put the registration? When I put it into `settings.py` which is the standard Django approach, it's overriden by djmoney. If I put registration in my models, there's no guarantees that it would be run after djmoney registration, the code would become dependent on module loading order, hence hard to understand and flaky.

I suggest removing automatical registration and instead asking the user to explicitly register djmoney serializer in `settings.py` if needed (e.g. if django-reversion is used in the project).

benjaoming:
I agree with this, if you can open a PR with the new behavior and put a good release note, it sounds good to me. We could target it for 2.2.

dbartenstein:
@emorozov: thanks for raising this issue. The [Django geojson serializer](https://docs.djangoproject.com/en/3.2/ref/contrib/gis/serializers/) also gets kicked out by django-money:

fhavret:
I created a PR to fix this issue, as it doesn't seem to be resolved.
Can I please have some feedback ? https://github.com/django-money/django-money/pull/729

PR #729, opened 2023-09-26
chore: don't register json serializer by default

fhavret:
Related to this issue https://github.com/django-money/django-money/issues/636, registering the serializer within `djmoney` package will override any customised json serializer provided by the user. The user should have the control of using the serializer provided by `djmoney` with Django `SERIALIZATION_MODULES` setting.


washeck:
I think this change requires documentation update and changelog entry.

benjaoming:
This will be a breaking change, so definitely :100: needs documentation, a changelog entry and a minor version bump.

fhavret:
Provided documentation, bumped version and added a changelog entry 👍 
Also a side question, why providing a custom deserializer but no custom serializer ?

docs/changes.rst
```diff
@@ -17,6 +17,13 @@ Changelog
 
 - (add entry here)
 
+`3.4` - unreleased
```
Stranger6667, line 20: I would leave this change to the commit which bumps the version and issues a release

Stranger6667: We have a template for this a few lines above, i.e the original intention was to fill this template (IRC)

fhavret: Updated 👍

Stranger6667: Thanks!

Stranger6667: Oh, I meant that the changelog entry should still be in this PR, but placed above what it used to be in the original commit in the templated block

fhavret: Ahh my bad ! Just updated it again

benjaoming (approved):
LGTM!

The one who does the next release (perhaps me? I'll write a TODO note) will notice this change in the changelog and know to bump the minor version instead of the patch version :+1:

Stranger6667 (approved):

PR #730, opened 2023-09-30
Add docs for validation forms with djmoney

leportella:
It took me a while to figure this out, so I decided to add an example to the docs :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -140,11 +140,13 @@ The ``balance`` field from the model above has the following validation:
 * Euros should be between 100 and 1000;
 * US Dollars should be between 50 and 500;
 
-Validation with forms
----------------------
+Constructing form data
+----------------------
 
-When using with the default ModelForm validation, it is expected that you should
-pass amount and currency as two separate fields:
+The default ``ModelForm`` class will use a form field (``djmoney.forms.fields.MoneyField``) that is constructed of two separate fields for amount and currency.
+
+If you need to feed data directly to such a form (for instance if you are writing a test case), then you need to pass amount and currency like this:
+
 
 
 .. code:: python
--- a/README.rst
+++ b/README.rst
@@ -165,7 +165,10 @@ If you need to feed data directly to such a form (for instance if you are writin
                 model = Product
                 fields = "__all__"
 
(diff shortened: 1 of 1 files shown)


rougeth (approved):

README.rst
```diff
@@ -140,6 +140,32 @@ The ``balance`` field from the model above has the following validation:
 * Euros should be between 100 and 1000;
 * US Dollars should be between 50 and 500;
 
+Validation with forms
+---------------------
+
+When using with the default ModelForm validation, it is expected that you should
+pass amount and currency as two separate fields:
```
benjaoming, line 147:
```suggestion
Constructing form data
----------------------

The default ``ModelForm`` class will use a form field (``djmoney.forms.fields.MoneyField``) that is constructed of two separate fields for amount and currency.

If you need to feed data directly to such a form (for instance if you are writing a test case), then you need to pass amount and currency like this:

```

README.rst
```diff
@@ -140,6 +140,32 @@ The ``balance`` field from the model above has the following validation:
...
+                default_currency='EUR'
+            )
+
+        # forms.py
+        class ProductForm(ModelForm):
+            class Meta:
+                model = Product
+                fields = "__all__"
+
+        ProductForm({'price_0': 10, 'price_1': 'EUR'})
```
benjaoming, line 166:
```suggestion
        # tests.py

        # construct the form in your test case
        form = ProductForm({'price_0': 10, 'price_1': 'EUR'})
```

README.rst
```diff
@@ -140,6 +140,32 @@ The ``balance`` field from the model above has the following validation:
...
+                max_digits=14,
+                decimal_places=2,
+                default_currency='EUR'
+            )
+
+        # forms.py
+        class ProductForm(ModelForm):
+            class Meta:
+                model = Product
+                fields = "__all__"
```
benjaoming, line 164:
```suggestion
                fields = ["price"]
```

benjaoming (approved):
I would like to make these changes to make it easier to understand what it's written for.... which I'm guessing is something with constructing tests.

benjaoming:
@leportella thanks for addressing the feedback, will check that the docs build works before merging. We've just recently enabled PR builds.

PR #737, opened 2023-10-30
Fix mypy issue in .pre-commit-config.yaml

benjaoming:
Fixes #736 

...maybe?

PR #738, opened 2023-11-20
Upper bound on pytest

benjaoming:


benjaoming:
I don't have time to fix this in a better way.. but it would help some other PRs to not fail their tests.

But in general, all test tools need to have bounds or be pinned. Otherwise maintaining anything back in time can become extremely tricky.

Stranger6667:
I am ok with this approach for the time being :) It looks like the only failing job is for the `mypy` issue which is going to be fixed with #737

Stranger6667 (approved):

PR #740, opened 2023-11-25
Updated CURRENCY_CHOICES to exclude currency "XXX" instead of DEFAULT_CURRENCY

### LINKED ISSUE #739: Excluding DEFAULT_CURRENCY from CURRENCY_CHOICES causes validation error
I noticed a curious piece of code in generating `CURRENCY_CHOICES` in `settings.py` of djmoney after I began running into odd validation issues after setting `DEFAULT_CURRENCY` to `USD` (currency object). It seems it is the default behavior to exclude the `DEFAULT_CURRENCY` from `CURRENCY_CHOICES`, although I find it confusing because 1) it raises a ValidationError when called `full_clean()` on a model which has the default value, since `DEFAULT_CURRENCY` *isn't* inside the accompanying `CurrencyField`'s choices, 2) it makes the `DEFAULT_CURRENCY` unselectable by the user on the frontend (I assume; I haven't actually tested this, but given model form factory constructs options via `choices`...). Is there a reason you've done this? On a *prima facie* review, there doesn't seem to be a good reason to do this, although I can't say for sure.

Here is the exact line of offending code in `settings.py`:
`CURRENCY_CHOICES = [(c.code, c.name) for i, c in CURRENCIES.items() if c != DEFAULT_CURRENCY]`

Naggafin:
Fixes #739

### DIFF
--- a/djmoney/settings.py
+++ b/djmoney/settings.py
@@ -4,7 +4,7 @@ from typing import Optional
 
 from django.conf import settings
 
-from moneyed import CURRENCIES, Currency
+from moneyed import CURRENCIES, XXX, Currency
 
 
 # The default currency, you can define this in your project's settings module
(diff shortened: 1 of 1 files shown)


benjaoming (approved):
Great work @Naggafin - wonderful that you found the explanation and made a comment about it!

If you want to add a release note about it for an upcoming release, that's great. The RST is a bit tricky, so I can also do that.

I will go ahead and merge this in a couple of days when I have to do a release, LMK if you have any additional comments @Stranger6667 - I'm think we can release this soon as a patch release. Strictly speaking, it should not break anything (if it does break something, it's likely a workaround for the bug that's fixed)

Stranger6667 (approved):

Naggafin:
Thanks :). Also sure, I can give it a shot. What do I need to do to add a release note? I don't know if this is too late already, also, but what do you think about the idea of an `EXCLUDED_CURRENCIES` setting? The thought came to me last night given the comparison check. There is of course a `PROJECT_CURRENCIES` setting variable already, but I figure having one that establishes an exclusive relationship vs an inclusive one might be more succinct for folks who want to have all currencies except certain ones. In this scenario, the `XXX` currency would always be a part of `EXCLUDED_CURRENCIES`, regardless of user specification, unless they specifically include it in `PROJECT_CURRENCIES`. What do you think?

Naggafin:
Understood! I'll make a separate PR for this at some point. How can I add a release note for what this PR addresses?

PR #741, opened 2023-11-27
Cleanup release notes

benjaoming:
Getting rid of a recurring annoyance :)

@Stranger6667 can you have a look at the Read the Docs builds? I think that PR builds are inactive, but I can't seem to reactivate them. I think it's the OAuth authentication between RTD and GitHub that's tied to another account from mine, perhaps yours?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,13 +36,15 @@ repos:
     -   id: flake8
         exclude: ^docs
 
--   repo: https://github.com/myint/rstcheck
-    rev: 3f92957
+-   repo: https://github.com/rstcheck/rstcheck
+    rev: v6.2.0
     hooks:
       - id: rstcheck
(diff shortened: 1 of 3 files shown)


Stranger6667:
I've updated GH settings, so the `pull_request` event is enabled now - my understanding is that RTD requires this event to build docs. Now, I guess that pushing here again should trigger the build on the RTD side.

benjaoming:
Should I try to re-push?

Stranger6667:
Yep, now everything should be fine

Stranger6667:
I just re-connected my GH account to RTD and the docs build seems to be reported to a PR of a different project, so, the next force push should trigger them here too.

Stranger6667:
Yep, it is here!

benjaoming:
Yes, indeed :clap:

PR #742, opened 2023-11-27
Test RTD

benjaoming:
Test if @Stranger6667's integration will trigger on this PR re: https://github.com/django-money/django-money/pull/741#issuecomment-1828058527


benjaoming:
Works!

PR #745, opened 2023-12-20
refactor: setup.py -> pyproject.toml

### LINKED ISSUE #659: setuptools should not be part of the dependencies
Setuptools is declared as a dependency here:

https://github.com/django-money/django-money/blob/98ddb3d9452fd374fb05b8a78f95145a38167e59/setup.py#L69

This is causing issues when installing the package as a dependency in a poetry project. It should be safe to just remove it from `install_requires`. And in fact I'd highly recommend migrating django-money to use pyproject.toml and poetry or some such, as that'll be a lot less of a headache to maintain. I can help with that if you like.

iacobfred:
Closes #659


Stranger6667:
I'd vote against having Poetry-specific content in `pyproject.toml`

Stranger6667:
Instead, I'd suggest specifying dependencies as described in the `pyproject.toml` [spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#dependencies-optional-dependencies).

benjaoming:
Nice work @iacobfred :pray: 

Agreeing with @Stranger6667, I don't see the current benefit of using Poetry. And even if switching to a tool like this, I would prefer hatch.. but then we have to do a lot of extra work to migrate current Tox-based environment management and CI. Which I think is too much for one changeset.

Would you be able to change to a version of pyproject.toml to be without Poetry and we just use `[project.optional-dependencies]` like @Stranger6667 linked?

benjaoming:
@Stranger6667 if we switch to hatch, we can get rid of the setup.py-based builder alltogether. Would you be good with that?

For instance like this:

https://github.com/django-wiki/django-wiki/blob/cf980084d222775458c976703972b7edca9aa07f/pyproject.toml#L67

and

https://github.com/django-wiki/django-wiki/blob/cf980084d222775458c976703972b7edca9aa07f/pyproject.toml#L84-L85

We could also use hatch to get rid of the Makefile and MANIFEST.in.

Stranger6667:
@benjaoming Yep!

browniebroke:
Is there an appetite to revisit this and potentially use [uv](https://docs.astral.sh/uv/)?

benjaoming:
@browniebroke yes :+1:

benjaoming:
This was now solved in https://github.com/django-money/django-money/pull/791

Thanks @jacobjove for starting the initiative :+1:

PR #747, opened 2024-01-26
Add magic methods float and int for Money class

kairkhan24:


benjaoming (changes requested):
Hi there!

It'd be good if you can state what you are trying to achieve here.

As you can see from the existing implementation of `__add__` and `__sub__`, we want to ensure that the currency is the same on both Money objects before any further arithmetic.

The definition of the parent class might help you here, as many arithmetic methods are already defined: https://github.com/py-moneyed/py-moneyed/blob/master/src/moneyed/classes.py#L158

kairkhan24:
> Hi there!
> 
> It'd be good if you can state what you are trying to achieve here.
> 
> As you can see from the existing implementation of `__add__` and `__sub__`, we want to ensure that the currency is the same on both Money objects before any further arithmetic.
> 
> The definition of the parent class might help you here, as many arithmetic methods are already defined: https://github.com/py-moneyed/py-moneyed/blob/master/src/moneyed/classes.py#L158

ok, i understood you, can we pass only float and int methods here?

benjaoming:
> And also is possible to write own CustomMoney inherited from Money class and pass it to MoneyField?

I don't know, I unfortunately haven't tried this and wouldn't know what pattern to follow to achieve it. You might have to subclass a lot of stuff.

benjaoming:
> Ok, i understood you, can we pass only float and int methods here?

I mean, if you want to improve and extend the functionality, that's great :partying_face: It was more to ensure that however it's done, it also includes the currency as you can see from existing methods.

benjaoming:
Closing because of missing feedback, but always welcome to reopen :+1:

PR #752, opened 2024-05-03
Map settings.BASE_CURRENCY to the base parameter for the openexchangerates.org backend

### LINKED ISSUE #751: openexchangerates.org base_currency parameter does not exist
According to the [docs](https://docs.openexchangerates.org/reference/set-base-currency)  `base_currency` should be mapped against the `base` parameter for the OpenExchangeRatesBackend

foarsitter:
This PR fixes #751 by setting `base` to `settings.BASE_CURRENCY`. 

To solve this problem in a more thorough manner, `get_params` should accept kwargs and the `base_currency` variable should passed to it. But this would require breaking changes.

### DIFF
--- a/djmoney/contrib/exchange/backends/openexchangerates.py
+++ b/djmoney/contrib/exchange/backends/openexchangerates.py
@@ -18,4 +18,4 @@ class OpenExchangeRatesBackend(SimpleExchangeBackend):
         self.access_key = access_key
 
     def get_params(self):
-        return {"app_id": self.access_key}
+        return {"app_id": self.access_key, "base": settings.BASE_CURRENCY}


benjaoming:
This looks like correct behavior :+1:

Have you tried it locally?

foarsitter:
Yes, applying this change to my working copy results in a 403 since I don't have a paid subscription. Setting  `BASE_CURRENCY = "USD"` results in the correct behavior. 

Sorry for the single quotes, always using pre-commit makes me forget that a thing as formatting exists.

djmoney/contrib/exchange/backends/openexchangerates.py
```diff
@@ -18,4 +18,4 @@ def __init__(self, url=settings.OPEN_EXCHANGE_RATES_URL, access_key=settings.OPE
         self.access_key = access_key
 
     def get_params(self):
-        return {"app_id": self.access_key}
+        return {"app_id": self.access_key, "base": settings.BASE_CURRENCY}
```
benjaoming, line 21:
```suggestion
        return {"app_id": self.access_key, "base": getattr(settings, "BASE_CURRENCY", "USD")}
```

benjaoming: It seems like current documentation doesn't require `BASE_CURRENCY` to be set but indicates that "USD" is the default.

> Backends return rates for a base currency, by default it is USD, but could be changed via BASE_CURRENCY setting. Open Exchanger Rates & Fixer supports some extra stuff, like historical data or restricting currencies in responses to the certain list. In order to use these features you could change default URLs for these backends:

benjaoming (changes requested):

djmoney/contrib/exchange/backends/openexchangerates.py
foarsitter: `update_rates` uses it as a default for a keyword argument. 
https://github.com/django-money/django-money/blob/c9749716fb9d98fe4bacb883798a2f28ee2c460a/djmoney/contrib/exchange/backends/base.py#L63

`BASE_CURRENCY` from djmoneny.settings does have a default, `BASE_CURRENCY` from django.conf.settings` doesn't so we are good here I guess.

benjaoming: Ay! You're right, thanks! Not sure why I missed that.

benjaoming (approved):

PR #753, opened 2024-05-03
Add Django 5.0 - Remove testing of non-supported Django 2.2, 3.2, 4.0, 4.1

benjaoming:
We can reduce some of the legacy support in turn of getting Django 5 in.

* Updated pre-commit
* Fixed some test cases for upcoming Django 6
* Upgraded to pytest 8.2


benjaoming:
@Stranger6667 were you okay with this? I should have waited for review, but am eager to have tests running again.

dreaquil:
Hi @benjaoming; Thanks for your work on adding Django 5.0. Just wanted to check in, it seems the [readme](https://github.com/django-money/django-money/blob/main/README.rst) does not mention django 5.0 support. Would be good to make this minor change.

PR #761, opened 2024-05-07
Rename CELERYBEAT_SCHEDULE to CELERY_BEAT_SCHEDULE

foarsitter:
Small adjustment to align the celery beat configuration with the new style naming convention. 

In the new convention the variable is called `beat_schedule` so it expects `CELERY_BEAT_SCHEDULE` when using celery with namespace='CELERY'.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -423,7 +423,7 @@ To set up a periodic rates update you could use Celery task:
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


benjaoming (approved):
Thanks for finding and fixing this!

Issue #768, opened 2024-05-31
The MoneyField from djmoney.contrib.django_rest_framework import MoneyField does not work anymore with custom serializer fields


gabn88:
Because of changes to the get_value function, the MoneyField only works for concrete fields, not for abstract fields (fields denoted with a @property).

Failing test:

```
class ProductSerializer(ModelSerializer):
    current_price = MoneyField(
        max_digits=10,
        decimal_places=2,
        default_currency='EUR',
        required=False,
        allow_null=True,
    )
    current_price_currency = serializers.CharField(source='price_A_currency', required=False)
```
with model
```
class Product(models.Model):
    @property
    def current_price(self):
        return ''
```

Caused by this function:

https://github.com/django-money/django-money/blob/f41d2f0c878a617bc671aa544089b9ca9b5c3064/djmoney/contrib/django_rest_framework/fields.py#L67

And more specifically this commit:
https://github.com/django-money/django-money/commit/c528c24c7e3b8c273ef85915881c23d652dc8386 

The error:
`FieldDoesNotExist`

Will provide a PR with a solution.

gabn88:
PR found here: https://github.com/django-money/django-money/pull/769 @benjaoming

PR #769, opened 2024-05-31
Update fields.py

gabn88:
Fixes the case where an error would be raised when using an abstract field as serializer MoneyField,

issue https://github.com/django-money/django-money/issues/768


benjaoming:
@gabn88 looks great :clap: and with a test case :100: 

What happens if the field really doesn't exist (not even as a property method)?

gabn88:
@benjaoming That case would also be fixed by this PR.

benjaoming:
The reason why this bug occurred in the first place is the lack of a test case.. so would be good to fix.

PR #772, opened 2024-07-30
fixes #768 where MoneyField breaks when using referencing properties …

### LINKED ISSUE #768: The MoneyField from djmoney.contrib.django_rest_framework import MoneyField does not work anymore with custom serializer fields

apjama:
…on the model in drf serializer. 

Issue here: https://github.com/django-money/django-money/issues/768

### DIFF
--- a/djmoney/contrib/django_rest_framework/fields.py
+++ b/djmoney/contrib/django_rest_framework/fields.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import FieldDoesNotExist
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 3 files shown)


benjaoming (approved):
Thanks for fixing this and adding tests, so we can keep it working :+1:

PR #776, opened 2024-10-02
Fix decompress for strings

rlaager:
If a string is passed, e.g. from a ModelAdmin /add/ query string, split it.  If no currency was provided, fall back to default_currency.

Without this, Django crashes.

### DIFF
--- a/djmoney/forms/widgets.py
+++ b/djmoney/forms/widgets.py
@@ -26,5 +26,8 @@ class MoneyWidget(MultiWidget):
         if value is not None:
             if isinstance(value, (list, tuple)):
                 return value
+            elif isinstance(value, str):
+                return (
+                    value.split(',') + [self.default_currency])[:2]
             return [value.amount, value.currency]
         return [None, self.default_currency]


benjaoming (changes requested):
Sorry about the late review! It's not so clear-cut for merging, would need some explanation and a test case.

benjaoming:
@rlaager thanks for illuminating :+1: do you have a documentation link or some kind of code segment where the API for this `/admin/app/model/add/?monthly_price=1%20USD` is elaborated?

rlaager:
> @rlaager thanks for illuminating 👍 do you have a documentation link or some kind of code segment where the API for this `/add/?field=value` syntax is elaborated?

I don't. The UI side of the admin interface (as opposed to the how to code `ModelAdmin`s) isn't particularly documented, from my experience.

I dug into the code some more.

[ModelAdmin.get_changeform_initial_data()](https://github.com/django/django/blob/c594574175e379fff356e274893d797f6e6a95fa/django/contrib/admin/options.py#L1767) is the entrypoint. It just takes `request.GET` and passes that (as a dict) as `initial` to the `ModelForm`.

Later, the widget is rendered. `BoundField.__str__()` calls `self.as_widget()`, which calls `self.value()` to pass as `value` to `widget.render()`. `Widget.render()` calls `MultiWidget.get_context()`. Since this is a MultiWidget, if the value is not already a list or tuple, it calls [self.decompress()](https://github.com/django/django/blob/c594574175e379fff356e274893d797f6e6a95fa/django/forms/widgets.py#L989).

So yes, as far as I can see, `decompress()` is the right place for this.

Major open questions:
1. Do you want to fix this? I'd argue you should, even if you don't think this is a popular use case, as crashes are always wrong.
2. What representation(s) for amount and/or amount & currency should be supported?

rlaager:
> > What representation(s) for amount and/or amount & currency should be supported?
> 
> That's a good question. I think I would go with the exact same string representation as the Money object that it maps to - so the string representation and `decompress` are each others' inverse. Something like:
> 
> ```
> assert Money("10.00", USD) == Money(*MoneyWidget.decompress(str(Money("10.00", USD))))
> ```

`str(Money("10.00", "USD"))` => `"$10.00"`

When we look at it in this direction, of course we know that this is USD. But if you are just given `"$10.00"`, you have no way of knowing if that is USD, CAD, AUD, etc.

Depending on the allowed `currency_choices`, though, it may not actually be ambiguous. We could be clever about that. That is probably good if people commonly use this for just one currency. If the field only allows "USD", then it's clear what "$10.00" means.

How about this, in `decompress()` where `isinstance(value, str)`:
```
currency = None
try:
    value, currency = value.split(' ', 1)
except ValueError:
    pass
amount = ''.join(x for x in value if x.isdigit())

if currency:
    for currency_choice, _ in self.currency_choices:
        if currency == currency_choice:
            break
    else:
        # A currency was specified, but it is not one of
        # the valid choices.
        # TODO: Maybe this should be a ValidationError?
        # TODO: Alternatively, should the value just be
        # TODO: ignored?
        raise ValueError(f"Unknown currency: {currency}")
else:
    for currency_choice, _ in self.currency_choices:
        if str(Money(amount, currency)) == value:
            if currency:
                # The value is something like $10.00,
                # which could be USD, CAD, etc.
                # TODO: If it's ambiguous, but matches the
                # TODO: default_currency, should that be
                # TODO: allowed?
                raise ValueError("Ambiguous currency")
            currency = currency_choice
    if not currency:
        currency = self.default_currency
```

With that, `"10.00"` would use the `default_currency` while `"$10.00"` would use whatever currency from `currency_choices` matches (as long as only one matches). Unambigous values like `"10.00 USD"` would work exactly as written (as long as the currency is in `currency_choices`).

rlaager:
Thoughts on this? If this sounds like a worthwhile direction, I can try to pick this up again.

benjaoming:
@rlaager I'm pretty sure that decompress is also used by other aspects of forms and fields. So I'm not sure if changing its behavior will be "for free".. but it seems like adding the `str` handling aspect isn't overlapping with anything else :+1: 

So I would say let's try it out and have some test cases for the various ways you've documented it to work.

I'm not sure if "$10.00" is needed. Is it in your use-case? If not, maybe just leave it out in order to start with the unambiguous string conversions.

rlaager:
> I'm not sure if "$10.00" is needed. Is it in your use-case?

It's not my use case.

My use case has been replaced by something else anyway.

I do think this is still worth fixing, as I take the approach that crashes are _always_ a bug. At a minimum, I suppose you could make `decompress()` simply eat the value and ignore it. That would fix the crash, making `/add/?field=value` a no-op.

PR #787, opened 2025-02-26
Add `backend` arg into `convert_money` function

heckad:


benjaoming (commented):
seems like a good change :+1:

benjaoming:
Sorry, there's a completely unrelated linting error:

```
djmoney/contrib/exchange/migrations/0001_initial.py:14: error: Cannot override class variable (previously declared on base class "Migration") with instance variable  [misc]
```

I guess we need to tell mypy to ignore migration files.

benjaoming:
@heckad yeah absolutely!

Issue #781, opened 2024-12-20
Typehints + bump requirements to Python 3.9 and Django 4.2


benjaoming:
It seems like we could make a lot of gains by introducing typehints.

This however requires to bump to Django 4.2 and Python 3.9.

* Django 4.2: Because it's the only currently supported version of Django and it supports Python 3.10
* Python 3.10: Because it supports the most conventional type annotations
* Remove old stubs

Any further modernizations are great, but not required.

Thoughts?

Related: #780

benjaoming:
Agreed @notypecheck 💯

benjaoming:
Now released as 3.6.0b3

PR #791, opened 2025-04-16
Migrate to pyproject.toml and uv

browniebroke:
I was planning to build on top of https://github.com/django-money/django-money/pull/745 but it turns out the base branch is quite out of date so started again from scratch.

Here is how I did it:

1. Used `hatch new --init` to automatically convert most of the file
2. Replaced mentions of hatch by `setuptools`
3. Replaced the `test` extra to a dependency-group
4. Replaced all pip install commands to use uv
5. Added uv plugins for pre-commit and tox
6. Added `uv.lock` file: I had to bump the `requires-python` from `>=3.7` to `>=3.8` as otherwise resolution failed:
	```
	  × No solution found when resolving dependencies for split (python_full_version == '3.7.*'):
	  ╰─▶ Because the requested Python version (>=3.7) does not satisfy Python>=3.8 and pytest>=8.2.0,<=8.2.2 depends on Python>=3.8, we can conclude that pytest>=8.2.0,<=8.2.2 cannot be used.
	```
    I assume that was overlooked when dropping Python 3.8, because there was no dependency resolution before...
7. Migrated the docs pip requirements to uv, [using a recipe I use elsewhere](https://github.com/browniebroke/django-remake-migrations/blob/4df3d1efd2fbaaae9b643d394d53d8154054be3e/.readthedocs.yml#L12-L17). The official docs have a slightly different variation now: https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv

Things I'm unsure and might need input from a maintainer:

1. There is a `Makefile` which uses some old commands, should I bother migrating it or is it no longer used?
2. Perhaps related to the previous question... How do you release currently? Is that automated [via this workflow](https://github.com/django-money/django-money/blob/main/.github/workflows/release.yml) or do I need to update some documentation?
3. ~Test against Django main are failing, which I assume will be fixed by https://github.com/django-money/django-money/pull/786?~ EDIT: this should be fixed


benjaoming:
If 3.5.4 works fine, I would love to release this as 3.6.

We could also consider trimming the Django support at bit as #781 suggests, but that can be a separate change (but nice to get in with a new minor version bump)

benjaoming:
@browniebroke thanks for doing the work of both implementing this and writing a tasty PR description :plate_with_cutlery:

pyproject.toml
```diff
@@ -1,7 +1,73 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = [
+    "Django>=2.2",
```
Stranger6667, line 41: Nit: looks like the lower bound is outdated here

pyproject.toml
```diff
@@ -1,7 +1,73 @@
...
 
+[project]
+name = "django-money"
+dynamic = ["version"]
+description = "Adds support for using money and currency fields in django models and forms. Uses py-moneyed as the money implementation."
+readme = "README.rst"
+license = { text = "BSD" }
+requires-python = ">=3.8"
+maintainers = [
+    { name = "Greg Reinbach", email = "greg@reinbach.com" },
```
Stranger6667, line 13: Even though I am barely active here, I think this list should be updated to reflect the current state of affairs with maintenance

pyproject.toml
```diff
@@ -1,7 +1,73 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = [
+    "Django>=2.2",
+    "py-moneyed>=2.0,<3.1",
+    "setuptools",
```
browniebroke, line 43: Is setuptools actually needed by the package or is it more of a build dependency?

pyproject.toml
```diff
@@ -1,7 +1,73 @@
 [build-system]
-requires = ["setuptools >= 40.6.0", "wheel"]
+requires = ["setuptools >= 75.3.2", "wheel"]
```
browniebroke, line 2: I've bumped this a bit to the latest version that support Python 3.8

.readthedocs.yaml
```diff
@@ -10,6 +10,12 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  commands:
+    - asdf plugin add uv
+    - asdf install uv latest
+    - asdf global uv latest
+    - uv sync --only-group docs --frozen
+    - uv run -m sphinx -T -b html -d docs/_build/doctrees -D language=en docs $READTHEDOCS_OUTPUT/html
```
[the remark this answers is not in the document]
browniebroke: Correct. I'm using this recipe in my own projects, which leverages the `docs` dependency groups ([PEP 735](https://peps.python.org/pep-0735/)). 

[RTD have a section about uv](https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv), but their solution still relies on `requirements.txt`.

setup.cfg
```diff
@@ -3,7 +3,7 @@ line_length = 120
 combine_as_imports = true
 known_django = django
 known_first_party = tests, djmoney, moneyed
-known_third_party = moneyed,pytest,rest_framework,setuptools,sphinx_rtd_theme
+known_third_party = moneyed,pytest,rest_framework,sphinx_rtd_theme
```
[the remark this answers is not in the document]
browniebroke: So I've hit a bit of a blocker on this:

- flake8 doesn't support `pyproject.toml`, so we'll have to either leave it here or move to `.flake8`
- I don't know what the `pep8` section is for
- `isort` supports `pyproject.toml`, but the `seed-isort-config` pre-commit hook doesn't. Although looking at the `seed-isort-config` repo, it's [archived and deprecated](https://github.com/asottile-archive/seed-isort-config)

Another option would be to move the whole thing to ruff, but that starts to feel out of scope of this PR. Thoughts?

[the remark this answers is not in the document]
benjaoming: > flake8 doesn't support pyproject.toml, so we'll have to either leave it here or move to .flake8

Hmm, we should use ruff instead. I wonder why that hasn't happened.

Maybe let's leave the setup.cfg with only the minimal set of configuration, and then when we introduce Ruff, we can get rid of the whole file :question:

.readthedocs.yaml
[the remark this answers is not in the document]
benjaoming: Alright, sounds good! I'm not for or against anything here, let's try it out! :pray: :+1:

pyproject.toml
benjaoming: That makes sense, thanks for doing that!

pyproject.toml
benjaoming: I don't think the package should depend on it. That might be an outdated thing. Does Git Blame show anything? Otherwise, let's get rid of it :+1:

benjaoming:
Thanks for your persistent work on this @browniebroke - seems very very close!!

pyproject.toml
browniebroke: Happy to add more maintainers here if you leave a suggestion 👍🏻

Stranger6667: Well, I think that @benjaoming and I (name = "Dmitry Dygalo", email = "dmitry at dygalo.dev") could be there, if there are no objections :)

pyproject.toml
browniebroke: I've tracked it down all the way back to https://github.com/django-money/django-money/commit/c01d29b4b2839976fd457a1e950ed5800150b315 (2012!)

I can see how it might have been needed back then, but I'm fairly sure it's no longer the case with the concept of build backend...

pyproject.toml
benjaoming: Yes, I think it's us two!

Benjamin Bach benjamin at overtag.dk

And I think that anyone with experience or specific visions to drive the project is always welcome to contact us :wink:

pyproject.toml
benjaoming: Agree!

pyproject.toml
benjaoming: You've already added me, nice! Thanks!

benjaoming:
Oh... juuuust one final question: Why is the uv.lock file present in a reusable package with loose dependencies?

browniebroke:
> Why is the uv.lock file present in a reusable package with loose dependencies?

Good question. It's perhaps a bit subjective, but I like to add it to ensure a consistent dev experience to contributors, with reproducible tests and builds. A key detail (that you may or may not be aware of) is that the lock file is NOT part of the package distribution in PyPI, so it won't affect users of the packages, only the folks who contribute to the project.

(I think I've read a blog post/guide making a good case for it, but I can't find it anymore)

benjaoming:
Let's try it out! I'm not sure how keeping it updated would work, usually we want tests to run on the latest dependencies allowed by loose pinning.

benjaoming (approved):
TY for the outstanding work :pray:

benjaoming:
I've noted on a TODO to get back to releasing this ASAP :+1:

benjaoming:
Unless @Stranger6667 beats me to it :smile:

browniebroke:
> Let's try it out! I'm not sure how keeping it updated would work, usually we want tests to run on the latest dependencies allowed by loose pinning.

I usually run a GitHub action workflow to update things regularly (~once a month) via a pull request.

Stranger6667:
> > Let's try it out! I'm not sure how keeping it updated would work, usually we want tests to run on the latest dependencies allowed by loose pinning.
> 
> I usually run a GitHub action workflow to update things regularly (~once a month) via a pull request.

I’d like to note, that Hypothesis has the same approach, it helps with addressing compatibility issues in a timely manner

benjaoming:
Does GHA have cron-like tasks? I know that Circle CI has and Travis (RIP) had

browniebroke:
> Does GHA have cron-like tasks? I know that Circle CI has and Travis (RIP) had

Yes it does. 

Here is how I do it in my repos: https://github.com/browniebroke/django-remake-migrations/blob/main/.github/workflows/upgrader.yml ([example PR](https://github.com/browniebroke/django-remake-migrations/pull/450))

And here is the reusable workflow if you prefer to have the actual steps: https://github.com/browniebroke/github-actions/blob/main/.github/workflows/uv-upgrade.yml

One thing to note is that I use a GitHub personal access token (PAT) to trigger the CI workflow on the PR. This is disabled by default in GitHub, as explained in [the create-pull-request action docs](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs). They offer a few other alternatives if you don't like the PAT appraoch.

PR #794, opened 2025-04-25
Update pre-commit hook versions and config

browniebroke:
- Update pre-commit hook versions
- Update config to target the minimum Python version currently supported (3.8)
- Run hooks to fix the code accordingly


benjaoming (approved):
Thanks!! :pray:

PR #810, opened 2025-11-19
Add getting a limit value from a callable

heckad:
Django's `BaseValidator` allow a callable for a `limit_value`. That is required for migrations. If we use a value from settings which can change, it leads to creating a new migration step.


heckad:
@benjaoming, any comments? This is a very important feature for us. Please look at this.

benjaoming:
@heckad this looks great to me :+1:

benjaoming (approved):
Tests look good and pass.

Will try to get a release note in about it.

PR #811, opened 2025-12-05
Ensure tox uses `uv.lock` file

browniebroke:
As per https://github.com/django-money/django-money/pull/809#issuecomment-3583474445, the tox setup currently doesn't use the `uv.lock` file to install test deps.

This PR changes the tox config to use dependencies group and `uv.lock` file to run the tests. This setup is pretty much copied from one of Adam Johnson's packages: https://github.com/adamchainz/django-cors-headers

Issue #793, opened 2025-04-25
Evaluate if uv.lock should stay


benjaoming:
https://github.com/django-money/django-money/pull/791 introduced uv.lock and we should revisit this choice to know if we really need a lock file in a distributed package or not.

@browniebroke has a suggestion on using GitHub Actions to keep the file updated: https://github.com/django-money/django-money/pull/791#issuecomment-2831085676

notypecheck:
It's should be ok to keep `uv.lock` there, it pins your regular and dev dependencies when developing the package, and when distributing users would install dependencies specified under `pyproject.toml/dependencies`, e.g. `typing-extensions>=4.12.2`.

benjaoming:
Hi @notypecheck 👋 what is the point of having pinned development if package users don't use them? (I'm asking the question out of genuine interest)

I think there needs to be some kind of pro/con list:

Pro:

* Developers can reproduce their environment

Con:

* Need to add extra chores, GitHub actions and noise around updating the .lock file. PRs, CI runs, merge conflicts etc.
* User issues may go unnoticed since users are not using pinned requirements

PR #816, opened 2026-03-10
Remove uv.lock file

browniebroke:
Added in https://github.com/django-money/django-money/pull/791 but I get the sense I never really managed to convinced the maintainer that it should be there in he first place. 

There is [a PR](https://github.com/django-money/django-money/pull/809) to keep it up to date but it hasn't been merged in several months so I think it's time revert this.

Refs #793
Closes #809
Closes #811


benjaoming:
Lot of things changing quickly, and it's nice to see both the fast development of uv (it's a fantastic tool!), but also nice to take some time to figure out what the ideal project management tool for Python maintainers will be in the future, given that uv is basically owned/bought by OpenAI.

I was never quite sure if the uv.lock file (but .lock files in general!) was ideal in a reusable Python package, I need like a list of pros/cons :smile:

browniebroke:
Fair enough. If we remove this file, there is basically no strings attached to uv and you could decide to switch back to pip which now supports dependency groups.

benjaoming (approved):

PR #819, opened 2026-04-11
Upsert rates instead of deleting and recreating them

### LINKED ISSUE #818: update_rates() should upsert rates instead of deleting and recreating them
The current implementation of `update_rates()` calls `backend.clear_rates()` before inserting new rates, [here](https://github.com/django-money/django-money/blob/main/djmoney/contrib/exchange/backends/base.py#L68). If the upstream backend returns a partial response and omits a currency it previously provided, that rate is permanently lost with no fallback, causing `MissingRate` exceptions somewhere down the line.

Suggested fix: replace the delete-then-insert pattern with an upsert - update rates that exist, create ones that are new, and leave stale ones untouched. This way a partial or degraded API response degrades gracefully instead of breaking the app.

2tunnels:
Fixes #818

### DIFF
--- a/djmoney/contrib/exchange/backends/base.py
+++ b/djmoney/contrib/exchange/backends/base.py
@@ -65,14 +65,16 @@ class BaseExchangeBackend:
         Updates rates for the given backend.
         """
         backend, _ = ExchangeBackend.objects.update_or_create(name=self.name, defaults={"base_currency": base_currency})
-        backend.clear_rates()
         params = self.get_params()
         params.update(base_currency=base_currency, **kwargs)
         Rate.objects.bulk_create(
             [
                 Rate(currency=currency, value=value, backend=backend)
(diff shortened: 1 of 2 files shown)


benjaoming (approved):
I agree with this change 👍 

I can see how the Rate handling could leave a lot to desire, for instance created/modified columns... and please feel free to open up additional issues for things you find handy 👍

benjaoming:
@2tunnels would you mind adding in a release note? :pray:

2tunnels:
@benjaoming done 🕺

benjaoming:
Amazing! Just making a slight tweak to the changelog, since "fixed" will convey that something trivially broken was simply fixed.. but this is a change in the behavior. Unlikely, but theoretically someone could be depending on the old behavior.

